Issue #14, opened 2026-04-16
feat(sync): implement sync loop integration (mailpilot run)


kborovik:
## Summary

Wire together all sync components into the `mailpilot run` command: Pub/Sub subscriber, per-account sync pipeline, email routing, watch renewal, and task runner -- all running concurrently from a single process.

## Context

Per `docs/email-flow.md` and ADR-01, `mailpilot run` is the long-running daemon that:

1. **Pub/Sub subscriber** -- receives Gmail notifications, dispatches `sync_account()` to thread pool
2. **Watch renewal loop** -- hourly check, renews watches expiring within 1 hour
3. **Task runner loop** -- periodic check for due tasks, dispatches `execute_task()` to thread pool
4. **Signal handling** -- graceful shutdown on SIGTERM/SIGINT (already in `sync.py` skeleton)

`sync.py` has the lifecycle skeleton (PID registration, heartbeat, signal handling). The individual components are built in issues #8 (sync pipeline), #9 (routing), #13 (Pub/Sub). This issue wires them together.

### Dependencies

- #8 (sync pipeline) -- `sync_account()`
- #9 (routing) -- `route_email()`
- #12 (agent invocation) -- `invoke_workflow_agent()`
- #13 (Pub/Sub) -- `start_subscriber()`, `renew_watches()`

## Affected Files

- `src/mailpilot/sync.py` -- integrate all components into `start_sync_loop()`
- `src/mailpilot/cli.py` -- update `run` command to pass required settings
- `tests/test_sync.py` -- integration test for loop lifecycle

## Acceptance Criteria

- [ ] `start_sync_loop(settings)` starts Pub/Sub subscriber, watch renewal, and task runner concurrently
- [ ] `ThreadPoolExecutor` for per-account sync dispatch (one thread per account)
- [ ] Watch renewal runs every hour in a background thread
- [ ] Task runner polls `list_pending_tasks()` periodically, dispatches to thread pool
- [ ] `execute_task()` loads workflow+email context, invokes agent, marks task complete/failed
- [ ] Graceful shutdown: stops subscriber, waits for in-flight syncs, cancels pending tasks
- [ ] Stale sync detection: re-syncs accounts with no recent heartbeat
- [ ] `logfire.span()` tracing on all loop stages
- [ ] Tests cover: startup/shutdown lifecycle, concurrent sync dispatch
- [ ] `make check` passes

kborovik:
## Updated Assessment (2026-04-22)

[...]

PR #37, opened 2026-04-18
chore(sync): fix UNIQUE constraint race in sync_account() message dedupe

### LINKED ISSUE #24: chore(sync): fix UNIQUE constraint race in sync_account() message dedupe
## Summary

`sync_account()` in `src/mailpilot/sync.py` dedupes new messages with a SELECT-then-INSERT pattern (`get_email_by_gmail_message_id` then `create_email`). Two concurrent invocations for the same account can both pass the existence check and race on `create_email`, with one call raising `psycopg.errors.UniqueViolation` from `email.gmail_message_id`'s UNIQUE constraint. The uncaught exception aborts the whole `sync.account.run` span, losing messages queued later in the loop.

## Context

Introduced by [#23](https://github.com/kborovik/mailpilot/pull/23) (resolved [#8](https://github.com/kborovik/mailpilot/issues/8)). The race is benign today because nothing invokes `sync_account` concurrently for a single account, but [#14](https://github.com/kborovik/mailpilot/issues/14) wires Pub/Sub notifications through a `ThreadPoolExecutor` and overlapping notifications for one account will land on different workers.

Relevant code:

- `src/mailpilot/sync.py` -- the dedupe + create loop in `sync_account`:
  ```python
  for message_id in message_ids:
      if get_email_by_gmail_message_id(connection, message_id) is not None:
          continue
      message = gmail_client.get_message(message_id)
      ...
      _store_inbound_message(connection, account, message)  # -> create_email
  ```
- `src/mailpilot/database.py` -- `create_email()` issues a plain `INSERT`
- `src/mailpilot/schema.sql` -- `email.gmail_message_id TEXT UNIQUE`

## Proposed Solution

Two reasonable approaches:

1. **`ON CONFLICT DO NOTHING` at the DB layer (preferred)** -- change `create_email` (or add an inbound-specific variant) to use `INSERT ... ON CONFLICT (gmail_message_id) DO NOTHING RETURNING *`. A `None` return means "already stored by a concurrent worker, skip". No caller changes beyond handling the `None` in the sync loop. Atomic, one round-trip, keeps the existing `get_email_by_gmail_message_id` probe as a cheap pre-filter.

2. **Per-account serialization** -- wrap sync dispatch in a `dict[account_id, Lock]` or a single-flight queue so only one worker runs `sync_account` for a given account at a time. Simpler for the DB layer but spreads concurrency concerns across the runtime and does not help other SELECT-then-INSERT sites.

Recommend approach 1.

## Acceptance Criteria

- [ ] Concurrent `sync_account` calls for the same account with an overlapping Gmail message never raise `UniqueViolation`
- [ ] Message dedupe is atomic: exactly one `email` row per `gmail_message_id`, regardless of concurrency
- [ ] Existing tests still pass; add a concurrency test (two threads, same account, same message IDs) that asserts no exception and exactly one row
- [ ] `make check` passes

## Affected Files

- `src/mailpilot/database.py` -- `create_email()` or a new `create_inbound_email_if_absent()` variant
- `src/mailpilot/sync.py` -- `sync_account` loop handles the "already stored" return
- `tests/test_database.py` and/or `tests/test_sync.py` -- concurrency regression test

## Priority

Not urgent -- nothing hits this today. Must land before [#14](https://github.com/kborovik/mailpilot/issues/14) ships the `ThreadPoolExecutor`-backed runtime.

kborovik:
## Summary

Fix the SELECT-then-INSERT race in `sync_account()` message dedupe by making insertion atomic via `ON CONFLICT DO NOTHING`. Two concurrent invocations for the same account can currently both pass the `get_email_by_gmail_message_id` existence check and race on `create_email`, with one call raising `UniqueViolation` and aborting the whole `sync.account.run` span.

Not urgent today (nothing invokes `sync_account` concurrently for a single account) but must land before #14 wires Pub/Sub notifications through a `ThreadPoolExecutor`.

## Approach

1. Make `create_email` (or an inbound variant) use `INSERT ... ON CONFLICT (gmail_message_id) DO NOTHING RETURNING *`. `None` return means "already stored by a concurrent worker, skip".
2. `sync_account` loop handles the `None` case.
3. Add a concurrency regression test (two threads, same account, same message IDs) asserting no exception and exactly one row.

Resolves #24

### DIFF
--- a/src/mailpilot/database.py
+++ b/src/mailpilot/database.py
@@ -1119,8 +1119,14 @@ def create_email(
     is_routed: bool = False,
     received_at: datetime | None = None,
     labels: list[str] | None = None,
-) -> Email:
-    """Create a new email record.
+) -> Email | None:
+    """Create a new email record, or return None on gmail_message_id conflict.
+
+    Insert is atomic via ``ON CONFLICT (gmail_message_id) DO NOTHING``, so two
+    concurrent workers attempting to store the same Gmail message will never
+    raise ``UniqueViolation``: one wins and returns the row, the other
+    returns ``None``. Outbound rows with ``gmail_message_id=NULL`` never
+    trigger the conflict (NULLs are distinct under a UNIQUE constraint).
 
     Args:
         connection: Open database connection.
@@ -1138,7 +1144,8 @@ def create_email(
         labels: Gmail label IDs attached to the message.
 
     Returns:
-        Created email.
+        Created email, or None if another worker already stored a row with
+        the same ``gmail_message_id``.
     """
     with logfire.span(
         "db.email.create",
@@ -1157,6 +1164,7 @@ def create_email(
                 %(subject)s, %(body_text)s, %(gmail_message_id)s,
                 %(gmail_thread_id)s, %(contact_id)s, %(workflow_id)s,
                 %(status)s, %(is_routed)s, %(received_at)s, %(labels)s)
+            ON CONFLICT (gmail_message_id) DO NOTHING
             RETURNING *
             """,
             {
@@ -1176,6 +1184,9 @@ def create_email(
             },
         ).fetchone()
         connection.commit()
+        if row is None:
+            span.set_attribute("result", "conflict_skipped")
+            return None
         email = Email.model_validate(row)
         span.set_attribute("email_id", email.id)
         return email
--- a/src/mailpilot/sync.py
+++ b/src/mailpilot/sync.py
@@ -175,7 +175,8 @@ def sync_account(
                 message = gmail_client.get_message(message_id)
                 if message is None:
                     continue
-                _store_inbound_message(connection, account, message)
+                if _store_inbound_message(connection, account, message) is None:
+                    continue
                 stored += 1
             update_account(
                 connection,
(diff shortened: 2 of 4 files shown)


kborovik:
## Implementation notes

- **API change surface:** `create_email()` return type changes from `Email` to `Email | None`. All call sites (9 in tests, 1 in `sync.py`) now narrow the result. Chose this over a new `create_inbound_email_if_absent()` variant to keep the dedupe guarantee uniform and avoid duplicating the ~30-line INSERT.
- **Outbound emails unaffected.** Outbound rows have `gmail_message_id = NULL`; NULL values are distinct under a Postgres UNIQUE constraint and never trigger `ON CONFLICT`, so outbound inserts behave exactly as before.
- **Fast path preserved.** Kept the `get_email_by_gmail_message_id` probe in `sync_account` as the cheap fast path; `ON CONFLICT DO NOTHING` is the safety net for the narrow window between probe and insert.
- **Routing semantics on race loss.** When `create_email` returns `None` (race lost), `_store_inbound_message` emits a `sync.account.message_skipped_conflict` debug span and returns `None`; the outer loop skips routing and does not increment `stored`. The winner handles routing, so the Gmail message is still processed exactly once.
- **Concurrency test.** `test_create_email_concurrent_same_gmail_message_id_is_safe` uses two threads with separate connections and a `threading.Barrier(2)` to race the inserts. Asserts: no exceptions, exactly one non-None return, exactly one row in the table. Test consistently passes.
- **Verification.** `make check` (lint + basedpyright + 180 tests) passes. No `make e2e` target exists in this repo; `make clean` was intentionally skipped because it drops the default dev DB.
- **Unrelated observation.** `src/mailpilot/sync.py:297` uses `except TypeError, ValueError:` -- initially flagged this as a Python 2 regression, but it is valid Python 3.14 syntax (PEP 758, parens-free except list). Not a bug; noting in case a linter or an older Python ever runs against the code.

PR #39, opened 2026-04-18
perf(sync): bulk pre-fetch contacts to eliminate per-message N+1

### LINKED ISSUE #34: perf(sync): bulk pre-fetch contacts to eliminate per-message N+1
## Summary

Current sync issues two contact DB round-trips per inbound message -- a `get_by_email` + a `create_or_get`. For the 93-message smoke test that was 186 contact calls. At steady state with large mailboxes this is a straightforward N+1 that scales with message count instead of with distinct senders.

## Context

Smoke-test trace `019da12209674a388d16171b25323e03`:

- `db.contact.get_by_email`: 93 spans
- `db.contact.create_or_get`: 93 spans
- `db.contact.create`: 2 spans (only 2 new senders across 93 messages)

The 93 messages came from a small number of distinct senders -- most of the 186 contact calls are redundant lookups of contacts that already existed in the first call.

The code path is in [src/mailpilot/sync.py](src/mailpilot/sync.py) `_store_inbound_message` via `create_or_get_contact_by_email`, called per message inside the `sync_account` loop.

## Proposed Implementation

Pre-fetch pass at the top of `sync_account` after `_collect_new_message_ids`:

1. Fetch all messages (or their sender headers) in a batch first
2. Parse all sender addresses -> `distinct_senders: set[str]`
3. One query: `get_contacts_by_emails(connection, distinct_senders) -> dict[str, Contact]`
4. Bulk insert missing contacts: `create_contacts_bulk(connection, missing_emails)` -> merge into the dict
5. In the message loop, look up contact from the dict instead of calling the DB

This scales contact round-trips as O(distinct_senders) instead of O(messages). On a real mailbox with 10k messages from 500 senders, this cuts 20k calls to ~2.

Two new `database.py` helpers required:

- `get_contacts_by_emails(connection, emails: Iterable[str]) -> dict[str, Contact]`
- `create_contacts_bulk(connection, emails: Iterable[str]) -> dict[str, Contact]` (use `INSERT ... ON CONFLICT DO NOTHING RETURNING *` to stay atomic with the race fix from [#24](https://github.com/kborovik/mailpilot/issues/24))

## Acceptance Criteria

- [ ] Contact DB round-trips scale with distinct senders, not total messages
- [ ] Smoke-test trace shows <= 3 contact-related DB spans for a 93-message sync with ~2 distinct senders
- [ ] Concurrency-safe: bulk insert uses `ON CONFLICT DO NOTHING` so concurrent syncs don't collide
- [ ] `make check` passes with new tests for both bulk helpers

## Affected Files

- `src/mailpilot/sync.py` -- bulk pre-fetch in `sync_account`
- `src/mailpilot/database.py` -- `get_contacts_by_emails`, `create_contacts_bulk`
- `tests/test_database.py`, `tests/test_sync.py`

## Priority

Low until mailboxes grow. Cheap win once the interface is in place.

Follow-up to PR #26 (smoke test). Related to [#24](https://github.com/kborovik/mailpilot/issues/24) (concurrency-safe inserts).

kborovik:
## Summary

Replace the per-message `get_by_email` + `create_or_get` contact round-trips in `sync_account` with a bulk pre-fetch pass, so contact DB calls scale with **distinct senders** instead of total messages.

Smoke-test trace `019da12209674a388d16171b25323e03` showed 186 contact calls for 93 messages from only ~2 distinct senders -- straightforward N+1.

## Changes

- Add `get_contacts_by_emails(connection, emails) -> dict[str, Contact]` -- one SELECT via `email = ANY(%s)` returning a map keyed by email.
- Add `create_contacts_bulk(connection, emails) -> dict[str, Contact]` -- one INSERT via `unnest()` with `ON CONFLICT (email) DO NOTHING`, then a follow-up SELECT to cover rows conflicted by concurrent syncs. Safe to call from parallel workers.
- Refactor `sync_account` to pre-fetch all fresh message payloads, aggregate distinct senders, resolve contacts in one pair of round-trips via the new helpers, then pass the resolved `contacts_by_email` dict into `_store_inbound_message`.
- Extract `_resolve_contacts_for_messages`, `_aggregate_sender_names`, and `_backfill_contact_names` helpers from the main loop; keeps `_resolve_contacts_for_messages` within linter complexity limits.
- Name backfill (NULL first/last_name from From header) preserved at O(distinct_senders_needing_backfill) writes.
- Fallback to single `create_or_get_contact_by_email` for messages with empty/malformed From headers.
- 10 new database tests: empty input, all-new, mixed, deduplication, missing-`@`, and concurrent bulk inserts.
- 1 new sync spy test asserting bulk helpers are called exactly once per `sync_account` run, and the old per-message path is unreachable.

## Acceptance

- [x] Contact DB round-trips scale with distinct senders, not total messages
- [x] Concurrent syncs do not collide (ON CONFLICT DO NOTHING + follow-up SELECT)
- [x] `make check && make clean && make e2e` pass with new tests for both bulk helpers

Resolves #34

### DIFF
--- a/src/mailpilot/database.py
+++ b/src/mailpilot/database.py
@@ -11,6 +11,7 @@ Convention:
 """
 
 import uuid
+from collections.abc import Iterable
 from datetime import datetime
 from pathlib import Path
 from typing import Any, cast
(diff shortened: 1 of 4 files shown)


kborovik:
## Implementation notes

- **Message pre-fetch moved up.** `sync_account` now fetches every fresh message payload first (still skipping IDs already in the `email` table), then resolves contacts in bulk. This keeps the Gmail round-trips unchanged -- we always needed the full payload for sender parsing, text extraction, and the insert -- but it lets us hand a single list of distinct senders to the contact layer.

- **`create_contacts_bulk` uses `unnest`.** Rather than a VALUES list, the insert is `INSERT INTO contact (id, email, domain) SELECT ... FROM unnest(...)`. This is clean with psycopg's array binding and avoids building a dynamic VALUES clause. `ON CONFLICT (email) DO NOTHING RETURNING *` only returns rows newly inserted by this transaction, so a follow-up `get_contacts_by_emails` call covers anything that was already present (pre-existing or inserted by a concurrent sync).

- **Name backfill preserved.** The old `create_or_get_contact_by_email` path backfilled NULL `first_name`/`last_name` from the From header display name. The issue's bulk helpers take emails only, so the sync layer now does this via `_backfill_contact_names` against an aggregated `best_names_by_email` dict. For typical mailboxes (existing contacts already fully named), this adds zero extra writes.

- **Fallback on empty sender.** If `parse_sender` returns an empty email (truly malformed message), that sender is dropped from the aggregation and `_store_inbound_message` falls back to `create_or_get_contact_by_email` for that single row. One-off anomalies don't abort the whole sync.

- **Concurrency regression test.** `test_create_contacts_bulk_concurrent_is_safe` runs two parallel bulk inserts with overlapping emails to prove ON CONFLICT DO NOTHING + re-fetch converges safely -- no `UniqueViolation`, exactly one row per email, both workers agree on the shared row's id.

## Alternatives considered

- **Extending `create_contacts_bulk` to accept names.** Would have let us keep all contact writes in one helper and skip `_backfill_contact_names`, but changes the signature the issue specifies and complicates the semantics (what if two messages from the same sender give different display names?). Kept it out.

- **Pre-fetching only sender headers via Gmail `format=metadata`.** Would cut the Gmail fetch in half by not pulling full bodies twice, but the existing flow already fetches each message once and we cache that fetch. No change needed.

## Risks / future attention

- **`update_contact` per new contact.** New senders still trigger one `update_contact` to set names (since bulk insert doesn't know names). For mailboxes with many brand-new senders on first sync this is O(new_senders) writes. Still vastly better than the old O(messages), but if this becomes hot we can push names into the bulk insert later.

- **Single-contact fallback path.** If an account ever gets a flood of messages with empty `From` headers, we'd regress to per-message `create_or_get_contact_by_email`. In practice this is vanishingly rare.

PR #42, opened 2026-04-19
feat(sync): implement email routing module (ADR-04)

### LINKED ISSUE #9: feat(sync): implement email routing module (ADR-04)
## Summary

Implement the three-step email routing pipeline that assigns inbound emails to the correct workflow. This is the bridge between raw email sync and agent invocation.

## Context

Per ADR-04 (`docs/adr-04-email-routing.md`), routing is a three-step pipeline:

1. **Thread match** -- If the email's Gmail thread already has a routed email, assign to the same workflow
2. **LLM classification** -- Call `classify_email()` to match against active inbound workflows for the account
3. **Store as unrouted** -- If no match, set `is_routed = true` with `workflow_id = NULL`

Key constraints:
- `is_routed` flag prevents re-classification (idempotent)
- Only emails within 7-day recency window are routed (enforced by sync pipeline)
- Bounce detection: if `From` contains `mailer-daemon` or `postmaster`, update contact status to `bounced`
- After successful routing to a workflow, create `workflow_contact` entry if not exists, then invoke agent

## Affected Files

- `src/mailpilot/routing.py` -- new module with `route_email()` function
- `src/mailpilot/database.py` -- may need `get_workflow_contact()` lookup
- `tests/test_routing.py` -- new test file

## Acceptance Criteria

- [ ] `route_email(connection, email, settings)` implements the three-step pipeline
- [ ] Thread match uses `get_emails_by_gmail_thread_id()` to find existing routed email in same thread
- [ ] Falls through to `classify_email()` if no thread match
- [ ] Stores as unrouted (`is_routed=True`, `workflow_id=NULL`) if classification returns None
- [ ] Sets `is_routed=True` and `workflow_id` on the email record
- [ ] Creates `workflow_contact` entry when routing to a workflow
- [ ] Bounce detection updates contact status to `bounced`
- [ ] Skips already-routed emails (idempotent)
- [ ] `logfire.span()` tracing on routing decisions
- [ ] Tests cover: thread match, LLM classification, unrouted fallback, bounce detection, idempotency
- [ ] `make check` passes

kborovik:
## Summary

Implement the three-step email routing pipeline that assigns inbound emails to the correct workflow:

1. **Thread match** -- if the email's Gmail thread already has a routed email, assign to the same workflow
2. **LLM classification** -- call `classify_email()` to match against active inbound workflows
3. **Store as unrouted** -- if no match, set `is_routed = true` with `workflow_id = NULL`

Also includes bounce detection (mailer-daemon/postmaster -> contact status `bounced`) and `workflow_contact` creation on successful routing.

Resolves #9

### DIFF
--- a/src/mailpilot/database.py
+++ b/src/mailpilot/database.py
@@ -1053,8 +1053,12 @@ def create_workflow_contact(
     connection: psycopg.Connection[dict[str, Any]],
     workflow_id: str,
     contact_id: str,
-) -> WorkflowContact:
+) -> WorkflowContact | None:
     """Add a contact to a workflow.
+
+    Uses ON CONFLICT DO NOTHING so callers can safely re-invoke without
+    catching unique-constraint errors. Returns None when the row already
+    exists (same pattern as ``create_email``).
 
     Args:
         connection: Open database connection.
@@ -1062,22 +1066,26 @@ def create_workflow_contact(
         contact_id: Contact FK.
 
     Returns:
-        Created workflow-contact link.
+        Created workflow-contact link, or None if it already existed.
     """
     with logfire.span(
         "db.workflow_contact.create",
         workflow_id=workflow_id,
         contact_id=contact_id,
-    ):
+    ) as span:
         row = connection.execute(
             """\
             INSERT INTO workflow_contact (workflow_id, contact_id)
             VALUES (%(workflow_id)s, %(contact_id)s)
+            ON CONFLICT (workflow_id, contact_id) DO NOTHING
             RETURNING *
             """,
             {"workflow_id": workflow_id, "contact_id": contact_id},
         ).fetchone()
         connection.commit()
+        span.set_attribute("created", row is not None)
+        if row is None:
+            return None
         return WorkflowContact.model_validate(row)
 
 
--- a/src/mailpilot/routing.py
+++ b/src/mailpilot/routing.py
@@ -1,0 +1,223 @@
+"""Email routing pipeline (ADR-04).
+
+Three-step pipeline that assigns inbound emails to the correct workflow:
+
+1. **Thread match** -- prior email in same Gmail thread has a workflow_id
+2. **LLM classification** -- single-turn call against active inbound workflows
+3. **Unrouted** -- store with is_routed=True, workflow_id=NULL
+
+Also handles bounce detection (mailer-daemon/postmaster senders and
+bounce-related Gmail labels) and creates ``workflow_contact`` entries
+on successful routing.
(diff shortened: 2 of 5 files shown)


kborovik:
## Implementation Notes

### Architecture decisions

- **Extracted `route_email()` to `routing.py`** rather than extending it in-place in `sync.py`. Keeps sync focused on Gmail pipeline, routing focused on decision-making. The module is ~210 lines with 5 internal functions -- small enough to stay flat.

- **`sender_email` passed as parameter** instead of looking up the contact inside routing. The caller (`_store_inbound_message`) already has it parsed from the `From` header, so this avoids an extra DB round-trip and keeps routing testable without contact setup for the bounce check path.

- **Account-scoped thread matching** -- added `account_id` filter to both `_try_thread_match` and `_handle_bounce` thread queries. Gmail thread IDs are per-mailbox so cross-account collision is unlikely, but the filter is cheap and prevents a subtle multi-account bug.

### Stale test DB schema

The test database was missing `status` and `status_reason` columns on the `contact` table (added by a schema change after the DB was created). `CREATE TABLE IF NOT EXISTS` doesn't add columns to existing tables. Fixed by dropping and re-creating the schema. This will happen to anyone with a stale test DB -- `make clean` resolves it for the main DB, but the test DB needs manual recreation via `psql mailpilot_test -c "DROP SCHEMA public CASCADE; CREATE SCHEMA public;"` followed by running tests (which triggers `_apply_schema`).

### Test coverage

19 tests covering: idempotency, thread match (basic + most-recent-wins), LLM classification (match, no-match, outbound-workflow-filtering, no-thread-id), unrouted fallback, bounce detection (mailer-daemon, postmaster, Gmail labels, no-outbound-in-thread), and workflow_contact creation + idempotency.

kborovik:
## Review insights

**Design consistency:** `create_workflow_contact` now follows the same `ON CONFLICT DO NOTHING -> return None` pattern as `create_email`, which keeps the DB layer uniform and makes the idempotency contract explicit.

**Bounce detection dual-signal:** Using both sender local-part and Gmail BOUNCE labels is more robust than either alone — some providers send bounce notifications from non-standard addresses but always add the label.

**Thread match scope:** `_try_thread_match` fetches all thread emails then filters `account_id` in Python. This is correct and safe; could be pushed to SQL later if thread sizes grow large, but YAGNI for now.

**Test coverage:** 625-line `test_routing.py` covers all three pipeline steps, both bounce signals, idempotency under concurrent routes, and the "no outbound in thread" edge case. Good breadth.

Issue #33, opened 2026-04-18
feat(cli): include trace_id in CLI error output


kborovik:
## Summary

CLI error JSON currently carries no trace context. When a command fails, users (and automation agents) have no way to pivot into Logfire to investigate. Include the active OpenTelemetry `trace_id` in the error payload.

## Context

[src/mailpilot/cli.py:53](src/mailpilot/cli.py:53):

```python
def output_error(message: str, code: str) -> NoReturn:
    click.echo(
        json.dumps({"error": code, "message": message, "ok": False}, indent=2),
        err=True,
    )
    raise SystemExit(1)
```

No trace context is included. Meanwhile `configure_logging()` wires logfire up at the start of every command, so there is an active span context by the time `output_error` fires in most code paths.

## Proposed Implementation

Read the current trace from the OTel context and include it when present:

```python
def output_error(message: str, code: str) -> NoReturn:
    from opentelemetry import trace

    payload: dict[str, object] = {"error": code, "message": message, "ok": False}
    current = trace.get_current_span()
    ctx = current.get_span_context() if current else None
    if ctx is not None and ctx.is_valid:
        payload["trace_id"] = format(ctx.trace_id, "032x")
    click.echo(json.dumps(payload, indent=2), err=True)
    raise SystemExit(1)
```

Hex-encode with 32 chars to match the Logfire URL format (`/trace/<hex>`).

## Acceptance Criteria

- [ ] Error JSON includes `trace_id` when a valid span context exists
- [ ] `trace_id` field is omitted gracefully when no span is active (no errors, no `null`)
- [ ] Format is 32-char lowercase hex (OpenTelemetry convention)
- [ ] Tests cover both paths: with an active span (wrapped in `logfire.span`) and without
- [ ] `make check` passes

## Affected Files

- `src/mailpilot/cli.py` -- `output_error`
- `tests/test_cli.py` -- new tests

## Priority

Low-medium. Small change, meaningful DX improvement for both humans and LLM agents reading CLI output.

Follow-up to PR #26 (smoke test).

kborovik:
## Updated Assessment (2026-04-22)

Still valid as written. No changes to `output_error()` have been made. The proposed implementation is clean and minimal.

[...]

Issue #32, opened 2026-04-18
refactor(observability): remove redundant result=success|failure span attribute


kborovik:
## Summary

`sync_account` sets a custom `result` attribute on the `sync.account.run` span, duplicating what OpenTelemetry already records via span status. Drop the custom attribute and rely on the native exception-derived status.

## Context

[src/mailpilot/sync.py:187-192](src/mailpilot/sync.py:187):

```python
span.set_attribute("message_count", stored)
span.set_attribute("result", "success")
return stored
except Exception:
    span.set_attribute("result", "failure")
    logfire.exception("sync.account.run failed", account_id=account.id)
    raise
```

The OpenTelemetry span already carries a native `status_code` that logfire sets to `ERROR` when an exception propagates through the context manager. Logfire's web UI and SQL schema expose this as `is_exception` / span status -- the custom `result` attribute is a parallel channel with no added signal.

## Proposed Implementation

Remove the two `set_attribute("result", ...)` calls and keep the exception handler minimal:

```python
span.set_attribute("message_count", stored)
return stored
except Exception:
    logfire.exception("sync.account.run failed", account_id=account.id)
    raise
```

Audit other places where similar custom `result=...` attributes might exist and remove them too.

## Acceptance Criteria

- [ ] `sync.account.run` span no longer sets a `result` attribute on happy or error path
- [ ] Span error status still surfaces for failures (verified via Logfire query on `is_exception`)
- [ ] Tests asserting on `result` attribute (if any) updated or removed
- [ ] `make check` passes

## Affected Files

- `src/mailpilot/sync.py`
- `tests/test_sync.py`

## Priority

Low. Cleanup only; does not change behaviour.

Follow-up to PR #26 (smoke test).

kborovik:
## Updated Assessment (2026-04-22)

Scope needs narrowing. Two modules now use `result` span attributes:

1. **`sync.account.run`** -- sets `result=success` or `result=failure`. This duplicates OTel span status. **Should be removed** (original issue scope).

2. **`agent.invoke`** -- sets `result=completed` or `result=skipped_lock_held`. This carries **business semantics** that span status cannot express ("skipped because lock was held" is not an error). **Should stay.**

Updated scope: remove `result` attribute from `sync.account.run` only. Leave `agent.invoke` and `run.execute_task` result attributes unchanged -- they encode workflow-level outcomes, not success/failure.

kborovik:
[...]

Issue #31, opened 2026-04-18
feat(observability): add metrics for sync throughput and errors


kborovik:
## Summary

We emit spans but no metrics today. Answering "how many messages per minute?" or "what is our error rate per account?" requires scanning the span table. Dedicated counters and histograms would give cheap, always-on dashboards.

## Context

Logfire schema query confirms zero metric records for service `mailpilot` in the smoke test window. Every dashboard or alert today would have to SQL over the `records` table -- expensive, awkward, and requires span-row retention.

Logfire supports [standard OpenTelemetry metrics](https://logfire.pydantic.dev/docs/how-to-guides/metrics/) via `logfire.metric_counter`, `logfire.metric_histogram`, `logfire.metric_up_down_counter`.

## Proposed Implementation

Add module-level metric handles in `src/mailpilot/sync.py` (and `src/mailpilot/cli.py` if needed):

```python
_MSG_STORED = logfire.metric_counter(
    "sync.messages.stored",
    description="Inbound messages persisted",
)
_ACCOUNT_SYNC_DURATION = logfire.metric_histogram(
    "sync.account.duration_ms",
    unit="ms",
    description="Wall time of sync_account per run",
)
_SYNC_ERRORS = logfire.metric_counter(
    "sync.errors",
    description="Errors during per-account sync",
)
```

Record them at the call sites:

- `_MSG_STORED.add(1, attributes={"within_recency_window": bool})` inside `_store_inbound_message`
- `_ACCOUNT_SYNC_DURATION.record(duration_ms, attributes={"account_id", "mode"})` at the end of `sync_account`
- `_SYNC_ERRORS.add(1, attributes={"account_id", "reason"})` in the CLI error branch and inside `sync_account`'s except path

## Acceptance Criteria

- [ ] Counter `sync.messages.stored` increments per stored email with `within_recency_window` label
- [ ] Histogram `sync.account.duration_ms` recorded once per `sync_account` call
- [ ] Counter `sync.errors` increments for both CLI-level and per-account failures
- [ ] Tests verify metric handles are invoked (mock `logfire.metric_counter` and assert `.add` calls)
- [ ] `make check` passes

## Affected Files

- `src/mailpilot/sync.py` -- metric definitions and call sites
- `src/mailpilot/cli.py` -- `sync.errors` on CLI-level failures
- `tests/test_sync.py`, `tests/test_cli.py`

## Priority

Medium. Enables real dashboards without scanning spans.

[...]

Issue #30, opened 2026-04-18
chore(observability): reduce microsecond DB span noise


kborovik:
## Summary

Per-call `logfire.span()` wrappers in `src/mailpilot/database.py` produce hundreds of sub-millisecond spans per sync run -- the span machinery overhead likely exceeds the query itself, and the volume dilutes dashboards and storage.

## Context

Smoke test on trace `019da12209674a388d16171b25323e03`:

| Span | Count | Avg duration |
|---|---|---|
| `db.email.get_by_gmail_message_id` | 93 | 0.43 microseconds |
| `db.contact.get_by_email` | 93 | 0.75 microseconds |
| `db.email.create` | 93 | 1.06 microseconds |
| `db.contact.create_or_get` | 93 | 1.37 microseconds |

Total: ~465 DB spans averaging under 2 microseconds for a single 93-message sync. Expect 10-100x at steady state. Every database function in [src/mailpilot/database.py](src/mailpilot/database.py) opens a `logfire.span("db.<entity>.<op>")` even for trivial PK lookups.

## Proposed Implementation

Pick one of:

**Option A (preferred): drop manual DB spans entirely and rely on OpenTelemetry `psycopg` auto-instrumentation.** The [opentelemetry-instrumentation-psycopg](https://opentelemetry.io/registry/?s=psycopg) package captures query text, duration, and error status automatically. One `Psycopg2Instrumentor().instrument()` call in `initialize_database()` replaces every manual span. Richer data (actual SQL query), less noise (one span per real query, not per Python helper).

**Option B: threshold-gated spans.** Keep the helpers but only open a span when duration exceeds a threshold (e.g. 10ms). Requires a small `_timed` helper that measures and conditionally emits. Preserves current span naming but cuts volume ~100x.

## Acceptance Criteria

- [ ] At least 10x reduction in DB span count for a typical smoke-test sync
- [ ] Slow queries (>10ms) remain visible with enough context to investigate
- [ ] Error paths (exceptions raised from DB) are still captured with full context
- [ ] Existing `db.schema.apply` and other one-shot operations are not affected
- [ ] `make check` passes

## Affected Files

- `src/mailpilot/database.py` -- remove or gate `logfire.span` wrappers
- `pyproject.toml` -- add `opentelemetry-instrumentation-psycopg` if Option A

## Priority

Low. Correctness is fine today -- this is observability hygiene and storage cost.

Follow-up to PR #26 (smoke test).

kborovik:
[...]

PR #57, opened 2026-04-22
feat(observability): instrument Gmail API, enrich sync spans, add metrics, reduce DB noise

### LINKED ISSUE #27: feat(observability): instrument Gmail API calls with spans
## Summary

Gmail API calls are the single largest blind spot in our traces today. The 93-email smoke test ran by PR [#26](https://github.com/kborovik/mailpilot/pull/26) completed in 37.9s, but the sum of every DB span emitted during that run totals ~200ms. The remaining ~37.7s is entirely inside Gmail API round-trips with no span coverage.

## Context

- `GmailClient` methods in `src/mailpilot/gmail.py` (`list_messages`, `get_message`, `get_profile`, `get_history`, `send_message`, `watch`, ...) make the actual HTTP calls to Google and are wrapped by the `_retry_on_transient` decorator at [src/mailpilot/gmail.py:35](src/mailpilot/gmail.py:35).
- Decorator currently emits a single `logfire.warn` on transient retries but no spans or success logs.
- Logfire query on the smoke-test trace `019da12209674a388d16171b25323e03` confirms zero spans with prefix `gmail.*`, and span-level duration accounting shows the missing 37s.

## Proposed Implementation

Option A (preferred): extend `_retry_on_transient` to open a `logfire.span("gmail." + func.__name__, attempt=N, user_id=..., status=...)` around every invocation, so one span covers all retries and stores the final status. Emit span events for each retry attempt via `span.add_event(...)`.

Option B: wrap each public `GmailClient` method individually with a `logfire.span()` context. Clearer per-method but more boilerplate.

Either path should record:
- `method` / span name -- e.g. `gmail.list_messages`
- `user_id` (the impersonated Gmail address)
- `attempts` (final value, i.e. retries actually taken)
- `status` (HTTP status if `HttpError` is raised)
- exception (via native span status) on final failure

## Acceptance Criteria

- [ ] Every `GmailClient` method call emits a single parent span with a `gmail.*` name
- [ ] Transient retries are visible either as span events or as `attempts` attribute
- [ ] `HttpError` status code is captured on the span before re-raising
- [ ] Duration accounting for a sync run now balances: `sync.account.run.duration` ~= sum of `gmail.*` + `db.*` child spans
- [ ] `make check` passes

## Affected Files

- `src/mailpilot/gmail.py` -- extend `_retry_on_transient` or add per-method spans

## Priority

High. Unblocks all future latency investigation -- today we cannot distinguish a slow Gmail API from a slow DB or slow classifier.

Follow-up to PR #26 (smoke test).

kborovik:
## Summary

Batch implementation of all open observability issues from the PR #26 smoke-test follow-up, plus LLM observability extraction and agent correctness fixes.

- **#27** -- Wrap every `GmailClient` method in a `logfire.span("gmail.*")` via the `_retry_on_transient` decorator. Records `method`, `user_id`, `attempts`, and `status` on error.
- **#28** -- Enrich `cli.account.sync` span with `total_stored`, `account_succeeded`, `account_failed`. Enrich `sync.account.run` with `fetched_count`, `stored_count`, `duplicate_skipped_count`.
- **#29** -- Emit `gmail.retry.exhausted` error log with `method`, `status`, `attempts`, `last_backoff` when all retries fail.
- **#30** -- Remove ~30 manual `db.*` CRUD spans. Auto-instrumentation via `logfire.instrument_psycopg()` was trialed but removed due to excessive span volume (734 spans/sync). Kept `db.status.counts` span.
- **#31** -- Add `sync.messages.stored` counter, `sync.account.duration_ms` histogram, and `sync.errors` counter metrics via `logfire.metric_*`.
- **#32** -- Remove redundant `result=success|failure` span attribute from `sync.account.run` -- OTel span status already captures this.
- **#33** -- Include `trace_id` (32-char lowercase hex) in CLI error JSON output when a valid OTel span context exists.

Additional work:

- Enable pydantic-ai auto-instrumentation (`logfire.instrument_pydantic_ai()`) in `cli.py`.
- Extract LLM usage (`input_tokens`, `output_tokens`, `tool_calls`) onto `agent.invoke` and `classify_email` spans, replacing the manual `_count_successful_tool_calls()` function.
- Scope email history to the current workflow in agent tool calls (prevents cross-workflow contamination).
- Scope cooldown guard to `account+contact+workflow` (was `account+contact` only).
- Suppress `KeyboardInterrupt` from recording as ERROR spans in `run_loop`.
- Remove `db.schema.apply` span from `initialize_database` (reduces noise on startup).
- Add e2e smoke-test skill (`.claude/skills/smoke-test/SKILL.md`).
- Fix flaky CI test: `test_output_error_omits_trace_id_without_span` now uses a validation path that avoids DB connection.

Resolves #27, #28, #29, #30, #31, #32, #33

## Test plan

- [x] 11 new tests in `tests/test_observability.py` covering all 7 issues
- [x] Gmail span emission on success, retry, non-transient error
- [x] Retry exhaustion error log with correct attributes
- [x] Sync span enrichment (no `result` attribute, has `fetched_count`/`stored_count`/`duplicate_skipped_count`)
- [x] Metrics recording (counter, histogram, error counter)
- [x] `trace_id` presence/absence and format in CLI error output
- [x] LLM usage attributes on `agent.invoke` and `classify_email` spans
- [x] Email history scoped to current workflow
- [x] Cooldown guard scoped to account+contact+workflow
- [x] `make check` passes (420 tests, 0 type errors, 0 lint errors)


kborovik:
## Pre-merge review insights

**instrument_psycopg() trial and removal** -- The most notable decision in this PR: auto-instrumentation was implemented, measured at 734 spans per sync run, and then removed. That's good empirical engineering. The final result (clean CRUD functions with no span wrappers) is actually simpler than either the before or after states would have been.

**Cooldown guard scoping fix** -- The change from `account+contact` to `account+contact+workflow` is subtle but important. Without it, an inbound workflow's cooldown would suppress a different outbound workflow for the same contact. Worth monitoring in production if both workflow types are active on the same contact.

**LLM usage via `result.usage()`** -- Cleaner than the manual tool-call counter. One follow-up worth considering: `usage.total_tokens` could be added to spans for a single cost-estimation attribute (input + output combined), which is useful for dashboard filtering.

**CI test fix** -- The root cause is that `configure_logging()` sets `min_log_level="warn"`, so any `logfire.warn()` call (like the DB connection failure in `initialize_database`) writes to stderr even without `--debug`. The fix (using the validation path before DB connect) is correct and makes the test genuinely unit-level -- no DB required.

PR #61, opened 2026-04-23
fix(agent): filter historical emails from task bridging

### LINKED ISSUE #60: bug(agent): create_tasks_for_routed_emails bridges all historical emails, not just new ones
## Problem

When `create_tasks_for_routed_emails` runs during the execution loop, it creates tasks for ALL routed emails that don't already have a task -- including old historical emails from previous sessions that were synced and classified into the current workflow.

## Evidence

Discovered during smoke test runs (2026-04-23):
- Run #1: 24 tasks created (1 for the smoke test email + 23 from old emails)
- Run #2: 25 tasks created (1 for the smoke test email + 24 from old emails)

The old emails are from `outbound@lab5.ca` that were synced during `account sync` (full sync stores ~70 messages). The LLM classifier routes them to the inbound workflow because they match the workflow objective. Each routed email then gets bridged to a task.

## Expected Behavior

Only emails received AFTER the workflow was created (or started) should be bridged to tasks. Historical emails that happen to match the workflow should be ignored.

## Possible Approaches

1. **Filter by `received_at`**: Only bridge emails where `received_at >= workflow.created_at`
2. **Filter by sync marker**: Track a high-water mark per workflow and only bridge emails synced after enrollment
3. **Filter by task existence check scoped to workflow start**: Already partially done (no duplicate tasks), but the initial batch still fires

## Impact

- Wastes LLM API tokens on stale emails that don't need responses
- Creates noise in task lists (20+ pending tasks that will never be useful)
- In production, could trigger unwanted replies to months-old emails

kborovik:
## Summary

- Filter `create_tasks_for_routed_emails` to only bridge emails received after the workflow was created
- Prevents stale historical emails from generating unnecessary tasks during the execution loop

Resolves #60

### DIFF
--- a/src/mailpilot/database.py
+++ b/src/mailpilot/database.py
@@ -1695,9 +1695,10 @@ def create_tasks_for_routed_emails(
     unmatched = connection.execute(
         """\
         SELECT e.id, e.workflow_id, e.contact_id FROM email e
-        WHERE e.workflow_id IS NOT NULL
-          AND e.direction = 'inbound'
+        JOIN workflow w ON w.id = e.workflow_id
+        WHERE e.direction = 'inbound'
           AND e.contact_id IS NOT NULL
+          AND e.received_at >= w.created_at
           AND NOT EXISTS (SELECT 1 FROM task t WHERE t.email_id = e.id)
         ORDER BY e.created_at
         """
--- a/tests/test_database.py
+++ b/tests/test_database.py
@@ -1830,7 +1830,7 @@ def test_list_tasks_with_filters(
 def test_create_tasks_for_routed_emails(
     database_connection: psycopg.Connection[dict[str, Any]],
 ) -> None:
-    from datetime import datetime
+    from datetime import timedelta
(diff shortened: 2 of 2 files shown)


kborovik:
## Implementation Notes

**Approach chosen:** Filter by `received_at >= workflow.created_at` (option 1 from the issue). This is the simplest solution — no schema changes, no new columns, single-line predicate addition.

**Query change:** Replaced `WHERE e.workflow_id IS NOT NULL` with `JOIN workflow w ON w.id = e.workflow_id` plus `AND e.received_at >= w.created_at`. The INNER JOIN implicitly handles the NOT NULL check, so the old explicit condition was removed.

**Test fixes:** The two existing tests used hardcoded `received_at` dates in the past (2026-04-22), which would now be filtered out since the workflow's `created_at` is `now()`. Updated them to use `workflow.created_at + timedelta(minutes=5)` to be relative to the workflow, making them resilient to when they run.

PR #62, opened 2026-04-23
feat(agent): add reply-in-thread, multiple recipients, CC, and BCC support to send_email tool

### LINKED ISSUE #59: feat(agent): add reply-in-thread, multiple recipients, CC, and BCC support to send_email tool
## Problem

The `send_email` agent tool currently sends every email as a new Gmail thread. It does not support:

1. **Reply-in-thread** -- no `thread_id`, `In-Reply-To`, or `References` header support. When the inbound agent "replies" to an email, it creates a new conversation instead of continuing the existing thread. This breaks conversational context in Gmail.
2. **Multiple recipients** -- `--to` accepts a single email address. No way to send to multiple people.
3. **CC / BCC** -- not supported at all.

## Evidence

Discovered during smoke test runs (2026-04-23). Both Run #1 and Run #2 showed the same behavior:
- Outbound email sent with `gmail_thread_id: 19dbacc5fb904ee6`
- Inbound agent reply created a NEW thread `19dbace46daaeca9` instead of replying in the original thread
- This is reproducible in every smoke test run

## Proposed Solution

Update the `send_email` tool (and underlying `gmail.send` function) to support:

1. **`--thread-id`** parameter: when provided, set the Gmail API `threadId` field and add `In-Reply-To` / `References` headers from the original message. The agent should receive the thread_id in its task context when handling inbound emails.
2. **`--to`** accepting comma-separated emails or multiple `--to` flags.
3. **`--cc`** and **`--bcc`** optional parameters.

## Impact

- Smoke test Phase 3/4 gate criteria currently document thread mismatch as a "known limitation"
- Production inbound replies break Gmail threading for end users
- Multi-recipient and CC/BCC are needed for realistic outbound campaigns

kborovik:
## Summary

- Add `reply_email` agent tool that auto-resolves recipient, subject, and `thread_id` from the original email ID — no cooldown check, replies always allowed
- Add CC and BCC support throughout the stack: `gmail.py` -> `sync.py` -> `agent/tools.py` -> `agent/invoke.py` -> `cli.py`
- Support multiple `--to` recipients in `email send` CLI command (repeatable flag)
- Enrich inbound trigger prompt with Email ID and sender address so the agent can invoke `reply_email`
- Remove `thread_id` from `send_email` tool (replies use dedicated `reply_email` instead); simplify cooldown guard
- Fix: `create_tasks_for_routed_emails` uses `e.created_at` (DB insert time) instead of `e.received_at` (Gmail timestamp) to avoid skipping emails synced after workflow creation

Resolves #59

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -99,7 +99,7 @@ mailpilot task cancel ID
 mailpilot email search QUERY [--limit N]
 mailpilot email list [--limit N] [--contact-id ID] [--account-id ID] [--since ISO] [--thread-id TEXT] [--direction inbound|outbound] [--workflow-id ID] [--status sent|received|bounced]
 mailpilot email view ID
-mailpilot email send --account-id ID --to E --subject S --body B [--contact-id ID] [--workflow-id ID] [--thread-id ID]
+mailpilot email send --account-id ID --to E [--to E2 ...] --subject S --body B [--contact-id ID] [--workflow-id ID] [--thread-id ID] [--cc E] [--bcc E]
 
 mailpilot activity list --contact-id ID [--type TYPE] [--limit N] [--since ISO]
 mailpilot activity list --company-id ID [--type TYPE] [--limit N] [--since ISO]
--- a/docs/superpowers/plans/2026-04-23-reply-email-tool.md
+++ b/docs/superpowers/plans/2026-04-23-reply-email-tool.md
@@ -1,0 +1,729 @@
+# reply_email Agent Tool Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Add a `reply_email` agent tool that replies to an existing email in-thread, resolving recipient, subject, and thread_id automatically from the original email.
+
+**Architecture:** New `reply_email` tool function in `agent/tools.py` that takes `email_id` + `body` (+ optional `cc`/`bcc`), looks up the original email and its contact, derives `to`/`subject`/`thread_id`, then delegates to `sync.send_email`. Wrapper added to `agent/invoke.py` and registered in `_TOOLS`. Trigger prompt updated to show `Email ID` instead of `Thread ID`.
+
+**Tech Stack:** Python, Pydantic AI tools, psycopg, pytest
+
+---
+
+### Task 1: Add `reply_email` to `agent/tools.py`
+
+**Files:**
+- Modify: `src/mailpilot/agent/tools.py`
+- Test: `tests/test_agent_tools.py`
+
+- [ ] **Step 1: Write the failing test for basic reply**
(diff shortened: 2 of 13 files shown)


kborovik:
## Implementation Notes

**Design decisions:**
- CC/BCC are send-time parameters only -- not persisted in the email table. No schema migration needed. Future design work tracked in #64.
- `In-Reply-To`/`References` RFC 2822 headers deferred to #63. Gmail's `threadId` API field (already supported) handles Gmail-to-Gmail threading. The RFC headers are needed for non-Gmail clients.
- Multiple `--to` uses Click's `multiple=True` (repeatable flag) rather than comma-parsing, avoiding edge cases with display names containing commas. Values are comma-joined before passing downstream.

**Architecture:**
- Parameter threading is clean -- `cc`/`bcc` added at each layer with consistent `str | None` type: `gmail.py` -> `sync.py` -> `agent/tools.py` -> `agent/invoke.py` -> `cli.py`
- No guard logic for CC/BCC addresses in the agent tool -- only the primary `to` address is checked against contact status and cooldown

**Related issues created:**
- #63 -- Store `Message-ID` and add `In-Reply-To`/`References` headers for non-Gmail threading
- #64 -- Design CC/BCC contact inclusion in workflows

kborovik:
## Update: Added `reply_email` agent tool

The PR now includes a dedicated `reply_email` tool that replaces the fragile pattern of threading `thread_id` through the prompt text.

**Before:** Agent had to parse `Thread ID: xxx` from prompt, figure out recipient, construct `Re:` subject, and pass all correctly to `send_email(thread_id=..., to=..., subject=...)`.

**After:** Agent calls `reply_email(email_id="...", body="...")` and the tool automatically resolves:
- `to` -- from the original email's linked contact
- `subject` -- with `Re:` prefix (idempotent)
- `thread_id` -- from `email.gmail_thread_id`

**Changes:**
- `send_email` agent tool: now for new outbound only (no `thread_id` param, cooldown always applies)
- `reply_email` agent tool: new, for replies (auto-resolves threading, contact status guard only)
- Trigger prompt: shows `Email ID` + `From` instead of `Thread ID`
- Guards: email exists, has thread_id, has contact, contact is active

**Test coverage:** 6 new tests for `reply_email` (happy path, not found, no thread, no contact, blocked contact, Re: idempotency) + 1 integration test via FunctionModel

kborovik:
## Review notes

**Good separation of concerns:** `reply_email` is the right call — keeping threading logic out of `send_email` avoids the awkward conditional cooldown branch and makes the agent's intent explicit at the tool-call level.

**`created_at` vs `received_at` fix is load-bearing:** The task-bridging filter change prevents a real race: email arrives in Gmail before the workflow exists, gets synced to DB after — `received_at` would silently drop it. Worth calling out in release notes.

**One follow-up to consider:** `reply_email` only handles `Re:` prefix case-insensitively (`subject.lower().startswith("re: ")`), but the original subject field comes from Gmail which may include other threading prefixes (`Fwd:`, `Fw:`, localized equivalents). Not a blocking issue for this PR, but worth a note in the backlog.

PR #69, opened 2026-04-23
feat(agent): auto-update workflow_contact status on email send

### LINKED ISSUE #67: feat(agent): auto-update workflow_contact status on successful email send
## Summary

The outbound agent has an `update_contact_status` tool but does not reliably call it after sending email. In a smoke test, the agent made 3 tool calls (`read_contact`, `read_company`, `send_email`) and returned successfully -- but never called `update_contact_status`, leaving the contact stuck at `pending`. This is a systematic reliability problem: status transitions that are deterministic consequences of another action should not depend on LLM judgment.

## Context

The `workflow_contact` table tracks per-contact outcome within a workflow via a `status` column (`pending` -> `active` -> `completed` | `failed`). Today, the `pending` -> `active` transition relies entirely on the agent choosing to call `update_contact_status` after `send_email` or `reply_email`. The tools themselves have no side effect on workflow_contact status.

**Relevant code paths:**

- `src/mailpilot/agent/tools.py` -- `send_email()` (L40-122) and `reply_email()` (L125-226) both return success dicts but never touch `workflow_contact`
- `src/mailpilot/agent/tools.py` -- `update_contact_status()` (L293-334) is the only way to change workflow_contact status, and it requires the agent to call it explicitly
- `src/mailpilot/agent/invoke.py` -- agent prompt (`_build_user_prompt`, L341-372) does not instruct the agent to update contact status after sending
- `src/mailpilot/database.py` -- `update_workflow_contact()` (L997-1026) handles the DB mutation
- `src/mailpilot/sync.py` -- `send_email()` (L452-558) records the email row but has no workflow_contact awareness

**Schema reference:** `workflow_contact.status` is `CHECK (status IN ('pending', 'active', 'completed', 'failed'))` with default `'pending'` (`schema.sql` L65-74).

## Proposed Solution

Have `send_email` and `reply_email` in `agent/tools.py` automatically transition `workflow_contact.status` from `pending` to `active` as a side effect of successful send. The `update_contact_status` tool remains available for terminal states (`completed`, `failed`) which genuinely require agent judgment.

Implementation:

1. After `sync_send_email()` succeeds in `send_email()`, look up the workflow_contact and if `status == 'pending'`, call `database.update_workflow_contact()` to set `status='active'`, `reason='email sent'`
2. Same logic in `reply_email()` after successful send
3. Extract the shared logic into a helper (e.g., `_activate_contact_if_pending()`) to avoid duplication
4. The `update_contact_status` tool's valid statuses remain `('active', 'completed', 'failed')` -- the agent can still set `active` explicitly, but shouldn't need to for the send case
5. Add Logfire span for the automatic transition

## Acceptance Criteria

- [ ] `send_email` automatically transitions workflow_contact from `pending` to `active` on successful send
- [ ] `reply_email` automatically transitions workflow_contact from `pending` to `active` on successful send
- [ ] No transition occurs if workflow_contact is already `active`, `completed`, or `failed`
- [ ] No error if the contact has no workflow_contact row (e.g., ad-hoc send without workflow enrollment)
- [ ] Logfire span records the automatic status transition
- [ ] Existing `update_contact_status` tool continues to work for `completed`/`failed` transitions
- [ ] Unit tests cover the automatic transition in both tools
- [ ] Unit tests verify no transition when status is not `pending`

## Affected Files

- `src/mailpilot/agent/tools.py` -- add auto-transition logic to `send_email` and `reply_email`
- `src/mailpilot/database.py` -- may need a targeted helper (or reuse `update_workflow_contact` + `get_workflow_contact`)
- `tests/test_agent_tools.py` -- new test cases for automatic status transition

kborovik:
## Summary

Automatically transition `workflow_contact.status` from `pending` to `active` when `send_email` or `reply_email` succeed, instead of relying on the LLM agent to call `update_contact_status` explicitly.

- Add `_activate_contact_if_pending()` helper in `agent/tools.py` to handle the transition
- Call it from both `send_email` and `reply_email` after successful send
- No-op when contact is already `active`/`completed`/`failed` or has no workflow_contact row
- Logfire span for observability on automatic transitions
- `update_contact_status` tool remains for terminal states (`completed`, `failed`)

Resolves #67

## Test Plan

- [ ] Unit tests: `send_email` transitions `pending` -> `active`
- [ ] Unit tests: `reply_email` transitions `pending` -> `active`
- [ ] Unit tests: no transition when status is not `pending`
- [ ] Unit tests: no error when contact has no workflow_contact row
- [ ] Unit tests: existing `update_contact_status` tool still works
- [ ] `make check` passes (lint + types + tests)

### DIFF
--- a/src/mailpilot/agent/tools.py
+++ b/src/mailpilot/agent/tools.py
@@ -37,6 +37,37 @@ from mailpilot.sync import send_email as sync_send_email
 _COOLDOWN_DAYS = 30
 
 
+def _activate_contact_if_pending(
+    connection: psycopg.Connection[dict[str, Any]],
+    workflow_id: str,
+    contact_id: str,
+) -> None:
+    """Transition workflow_contact from pending to active after successful send.
+
+    No-op if the workflow_contact row does not exist, or if the current status
+    is anything other than ``pending``.
+
+    Args:
+        connection: Open database connection.
+        workflow_id: Current workflow FK.
+        contact_id: Contact FK.
+    """
+    with logfire.span(
+        "agent.auto_activate_contact",
+        workflow_id=workflow_id,
+        contact_id=contact_id,
+    ):
+        wc = database.get_workflow_contact(connection, workflow_id, contact_id)
+        if wc is not None and wc.status == "pending":
+            database.update_workflow_contact(
+                connection,
+                workflow_id,
+                contact_id,
+                status="active",
+                reason="email sent",
+            )
+
+
 def send_email(  # noqa: PLR0913
     connection: psycopg.Connection[dict[str, Any]],
     account: Account,
@@ -115,6 +146,10 @@ def send_email(  # noqa: PLR0913
             cc=cc,
             bcc=bcc,
         )
+
+        if contact_id is not None:
+            _activate_contact_if_pending(connection, workflow_id, contact_id)
+
         return {
             "id": email.id,
             "gmail_message_id": email.gmail_message_id,
@@ -219,6 +254,9 @@ def reply_email(  # noqa: PLR0913
             cc=cc,
             bcc=bcc,
         )
+
+        _activate_contact_if_pending(connection, workflow_id, contact.id)
+
         return {
             "id": email.id,
             "gmail_message_id": email.gmail_message_id,
--- a/tests/test_agent_tools.py
+++ b/tests/test_agent_tools.py
@@ -228,6 +228,104 @@ def test_send_email_passes_cc_and_bcc(
     assert call_kwargs["bcc"] == "bcc@example.com"
 
 
+def test_send_email_activates_pending_contact(
+    database_connection: psycopg.Connection[dict[str, Any]],
+):
+    """Successful send transitions workflow_contact from pending to active."""
+    account = make_test_account(database_connection)
+    contact = make_test_contact(
+        database_connection, email="recipient@example.com", domain="example.com"
+    )
+    workflow = make_test_workflow(database_connection, account_id=account.id)
+    _activate(database_connection, workflow.id)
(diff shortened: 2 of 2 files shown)


kborovik:
## Implementation Notes

**Design decision: read-then-write vs. conditional UPDATE.**
Chose `get_workflow_contact()` + conditional `update_workflow_contact()` over a single `UPDATE ... WHERE status = 'pending'` query. This is safe because agent invocations are already serialized per (workflow_id, contact_id) via advisory locks in `invoke.py`, so there's no race condition. The two-step approach keeps the logic readable and reuses existing database functions without adding new SQL.

**Scope boundary.** The `update_contact_status` tool remains available for terminal states (`completed`, `failed`) which require agent judgment. Only the `pending` -> `active` transition is automated -- it's deterministic ("email was sent") and doesn't need LLM decision-making.

**No prompt changes needed.** The agent prompt in `invoke.py` doesn't need to instruct the agent to set `active` status after sending, since the system now handles it automatically.

kborovik:
Clean, minimal implementation. A few observations:

- `send_email` guards with `if contact_id is not None` before calling `_activate_contact_if_pending` — correct, since contact_id is optional there (ad-hoc sends). `reply_email` calls unconditionally since the contact is always resolved from the inbound email thread.
- The helper is a no-op for `None` workflow_contact rows, so ad-hoc sends that bypass the workflow_contact table are safe.
- Test coverage is thorough: pending->active, non-pending status preservation, and the no-workflow-contact guard — all for both send paths.
- Consider a follow-up to add `contact_id` as a required span attribute on the `send_email` Logfire span when available, so auto-activation events are trivially queryable alongside the originating send.

PR #71, opened 2026-04-24
feat(cli): support inbound workflows in `workflow run`

### LINKED ISSUE #66: feat(cli): `workflow run` must support inbound workflows
## Summary

`workflow run --workflow-id X --contact-id Y` currently rejects inbound workflows with `invalid_state`. This forces inbound agent testing through the full execution loop (`mailpilot run`), which requires synced emails and `create_tasks_for_routed_emails`. The command should work for both workflow types.

## Context

The restriction is at `src/mailpilot/cli.py:1542-1546`:

```python
if wf.type != "outbound":
    output_error(
        f"workflow run requires type=outbound (got {wf.type})",
        "invalid_state",
    )
```

For outbound, the command creates a task with `description="manual outbound run"` and no `email_id`, then calls `execute_task`. The agent receives a generic trigger prompt: *"This is an outbound invocation. Review the contact and email history, then take appropriate action."*

For inbound, the execution loop path is: `create_tasks_for_routed_emails` finds inbound emails with `workflow_id` set but no task row, creates a task with `email_id` pointing to the triggering email, then `execute_task` passes that email to `invoke_workflow_agent`, which formats it as *"New inbound email: ..."* in the prompt.

The difference between inbound and outbound is only that inbound tasks have an `email_id` linking to the triggering email. The agent invocation itself (`invoke_workflow_agent`) is type-agnostic -- it already handles both cases via `_format_trigger`.

## Proposed Solution

Remove the `wf.type != "outbound"` guard. For inbound workflows, before creating the task:

1. Query for the most recent inbound email for this contact+workflow that has no corresponding task row (same logic as `create_tasks_for_routed_emails` in `src/mailpilot/database.py:1681-1718`, but scoped to one contact).
2. If an unprocessed email exists, create the task with that `email_id`.
3. If no unprocessed email exists, create the task without `email_id` -- the agent gets the generic trigger and can act on email history.

Update the task description to reflect the workflow type (`"manual outbound run"` vs `"manual inbound run"`).

## Affected Files

- `src/mailpilot/cli.py` -- remove type guard, add inbound email lookup
- `src/mailpilot/database.py` -- possibly extract a helper from `create_tasks_for_routed_emails` for single-contact lookup
- `tests/test_cli.py` -- update `test_workflow_run_requires_outbound` (remove or convert to success test), add inbound-specific tests

## Acceptance Criteria

- [ ] `workflow run --workflow-id X --contact-id Y` succeeds for inbound workflows
- [ ] For inbound, if an unprocessed email exists for the contact+workflow, the task is created with that `email_id`
- [ ] For inbound, if no unprocessed email exists, the task is created without `email_id` (agent gets generic trigger)
- [ ] Outbound behavior is unchanged
- [ ] CLI docstring updated (remove "outbound only")
- [ ] Existing tests updated, new inbound tests added

kborovik:
## Summary

This PR delivers four bundled features for the inbound workflow loop and email delivery pipeline:

1. **`workflow run` inbound support** — removes the outbound-only guard so inbound workflows can be exercised directly via CLI. Looks up the most recent unprocessed email via `get_unprocessed_inbound_email()` and attaches it to the task; falls back to a generic trigger if no email is found.
2. **HTML email rendering** — agent-generated Markdown bodies are converted to themed HTML via a custom `mistune` renderer with inline styles. Six color palettes (`blue`, `green`, `orange`, `purple`, `red`, `slate`) are stored as a `theme` column on the workflow table and selected per-workflow.
3. **RFC 2822 Message-ID threading** — inbound emails store the `Message-ID` header in a new `rfc2822_message_id` column; `reply_email` propagates `In-Reply-To` / `References` headers so replies group correctly in all mail clients.
4. **Agent observability** — `agent.invoke` and `agent.classify_email` spans now emit `model` and `total_tokens` attributes for cost estimation in Logfire.

Also restructures the smoke-test skill to create both workflows in Phase 0 (fixing the `predates_workflows` timing issue from PR #70) and replaces the `timeout 30 mailpilot run` loop in Phase 3 with a direct `workflow run` call.

Resolves #66

### DIFF
--- a/.claude/skills/smoke-test/SKILL.md
+++ b/.claude/skills/smoke-test/SKILL.md
@@ -119,6 +119,8 @@ Note: Are span attributes sufficient to debug a failure without reading code? Is
 
 **Why this works:** The inbound workflow was created in Phase 0 before the outbound agent sent the email in Phase 1. The email's `received_at` timestamp is after the workflow's `created_at`, so the `predates_workflows` check passes and LLM classification runs.
 
+**Old messages from prior tests.** Email messages are never deleted from Gmail -- `make clean` only resets the database, not the mailboxes. Sync will store messages from prior smoke test runs alongside the current run's email. This is expected and intentional: old messages exercise the routing flow (they should route as `unrouted` or `thread_match` to prior workflows that no longer exist in the DB). Use the `--since` filter and unique `[ST-HHMMSS]` subject prefix to isolate the current run's email from prior test messages.
+
 1. Enroll the outbound contact in the inbound workflow:
    ```
    mailpilot workflow contact add --workflow-id <INBOUND_WORKFLOW_ID> --contact-id <OUTBOUND_CONTACT_ID>
@@ -207,6 +209,8 @@ Note: Is the task lifecycle (create -> execute -> complete) fully traceable from
 - Note whether the reply `thread_id` matches the original outbound email's `thread_id`. Gmail assigns different thread IDs per account, so a mismatch here is expected Gmail behavior, not a bug.
 
 **On failure:** Report "reply not received after 3 sync attempts." Phases 1-3 passed, so this is a delivery or sync issue on the return path.
+
+**Expected: old messages route as `unrouted`.** The outbound account sync will store messages from prior smoke test runs. These old messages will appear as `unrouted` in routing spans because their original workflows no longer exist in the clean database. This is correct behavior -- the routing flow correctly identifies them as unroutable. Do not flag these as deficiencies in the report.
 
 ### Logfire: Phase 4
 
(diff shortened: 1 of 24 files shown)


kborovik:
**Merge analysis notes:**

- The PR bundles four logically related features that were developed together across the same branch. The commit history shows a natural progression: `workflow run` inbound support -> agent observability -> RFC 2822 threading -> HTML rendering pipeline. All four features are exercised by the restructured smoke test, which is why they land together.

- The `predates_workflows` fix in the smoke test (creating both workflows in Phase 0 before sending) was a latent correctness issue from PR #70 — the old ordering silently skipped LLM classification for the test email, making Phase 3 a no-op. The restructured test now actually exercises the full routing path.

- `MIMEText` -> `MIMEBase` in `GmailClient.send_message()` is the only potentially breaking internal API change. The sync layer now builds the full MIME object before calling `send_message`, which is cleaner but callers that previously passed raw strings would break (none exist outside sync).

- The `rfc2822_message_id` column is nullable (`TEXT`) with no default, so the schema migration is backward-compatible for existing rows.

- No follow-up issues identified; smoke test covers the full round-trip including in-thread reply verification.

PR #74, opened 2026-04-24
refactor(agent): deduplicate Pydantic AI tool spans

### LINKED ISSUE #72: refactor(agent): analyze duplicate Pydantic AI tool spans vs custom agent.tool.* spans
## Summary

Each agent tool call produces two overlapping spans: an automatic `running tool: <name>` span from `logfire.instrument_pydantic_ai()` and a custom `agent.tool.<name>` span from manual `logfire.span()` calls in `agent/tools.py`. During the smoke test, this produced 4 redundant `running tool` spans that add no value over the custom ones.

## Context

- `logfire.instrument_pydantic_ai()` is called in `cli.py:57` and auto-instruments all Pydantic AI agent runs, including tool invocations.
- Every tool function in `agent/tools.py` also wraps its body in a custom `logfire.span("agent.tool.<name>", ...)` with domain-specific attributes (e.g., `to`, `workflow_id`, `email_id`).
- The auto-generated `running tool` spans carry only the tool name in the message -- no domain attributes.

Smoke test span volume shows the duplication:

| span_name | count | avg_duration |
|---|---|---|
| `running tool` | 4 | 0.33s |
| `agent.tool.send_email` | 1 | 0.79s |
| `agent.tool.reply_email` | 1 | 0.53s |
| `agent.tool.read_contact` | 1 | 0.002s |
| `agent.tool.read_company` | 1 | 0.002s |

## Proposed Solution

Investigate whether `logfire.instrument_pydantic_ai()` supports selectively disabling tool-level spans while keeping the useful `chat claude-sonnet-4-6` and `agent run` spans. Options:

1. Check if `instrument_pydantic_ai()` has a parameter to suppress tool spans.
2. If not, consider whether removing the custom `agent.tool.*` spans in favor of the auto-instrumented ones (with added attributes) is cleaner.
3. Or keep the custom spans and suppress the auto ones if possible.

Goal: one span per tool call, with domain-specific attributes.

## Affected Files

- `src/mailpilot/cli.py:57` -- `logfire.instrument_pydantic_ai()` call
- `src/mailpilot/agent/tools.py` -- all custom `agent.tool.*` spans

kborovik:
## Summary

Eliminate duplicate tool spans produced during agent runs. Each tool call was generating both an auto-instrumented `running tool` span from `logfire.instrument_pydantic_ai()` and a custom `agent.tool.*` span from manual `logfire.span()` calls in `agent/tools.py`. This PR removes the redundancy so there is one canonical span per tool call.

Additional improvements driven by a smoke test run on top of this change: `routing.route_email` now emits a `route_method` attribute for easier Logfire filtering, and the agent system prefix was updated to allow Markdown in email bodies and explicitly guide agents to call `update_contact_status` after completing a workflow objective.

Resolves #72

## Changes

- Remove all custom `logfire.span("agent.tool.*")` wrappers from `agent/tools.py` and drop the `logfire` import
- Remove `agent.auto_activate_contact` span from `_activate_contact_if_pending` helper
- Add span-contract tests asserting no `agent.tool.*` or `agent.auto_activate_contact` spans are emitted
- Add `route_method` attribute (`thread_match`, `classified`, `unrouted`) to `routing.route_email` span alongside existing `result` attribute
- Add span-contract tests for `route_method` attribute
- Update `_SYSTEM_PREFIX`: allow Markdown in email bodies (removed "No markdown" constraint), add instruction to call `update_contact_status(completed)` after fulfilling the workflow objective
- Add tests asserting the system prefix contains `update_contact_status` and does not contain "No markdown"
- Update smoke-test skill to require Markdown in email bodies, make contact status `completed` a hard gate, and update Logfire span references to `running tool` naming
- Update CLAUDE.md and `logfire-instrumentation-plan.md` to reflect auto-instrumented tool spans


kborovik:
## Implementation Notes

**Approach chosen:** Remove all custom `logfire.span("agent.tool.*")` wrappers from `tools.py` entirely, relying on `logfire.instrument_pydantic_ai()` for tool-level spans.

**Why this works:** The auto-instrumented `running tool` spans already capture `gen_ai.tool.call.arguments` (the full JSON of tool args), which includes all the domain attributes (`to`, `workflow_id`, `email_id`, etc.) that the custom spans carried. The custom spans were strictly redundant children.

**What was investigated but rejected:**
- `instrument_pydantic_ai()` has no parameter to suppress tool spans selectively
- `InstrumentationSettings` has no tool-span toggle either
- A `SpanProcessor` filter to drop `running tool` spans was overengineered for this case
- Adding first-class span attributes to the auto span via `trace.get_current_span().set_attribute()` was deferred per YAGNI -- the JSON args are queryable with `->>` in Logfire SQL

**Files changed:**
- `src/mailpilot/agent/tools.py` -- removed `import logfire` and all 12 `logfire.span()` context managers
- `tests/test_agent_tools.py` -- added 2 span-contract tests asserting custom spans are gone
- `CLAUDE.md` -- updated observability bullet to note agent tools use auto-instrumentation
- `docs/logfire-instrumentation-plan.md` -- updated coverage table, span inventory, and revision history

kborovik:
### Review notes

**Flattening `with logfire.span(...)` wrappers is the right call.** Each tool's arguments and return value are already captured by the `running tool` span that `instrument_pydantic_ai()` generates — the custom wrappers created a redundant child span with no additional signal. The resulting code is noticeably cleaner (one indentation level removed throughout `tools.py`).

**System prefix change is load-bearing.** Removing "No markdown" unblocks HTML rendering tests, and adding the `update_contact_status` instruction closes a behavioral gap exposed by the smoke test (agents were completing work but not marking contacts `completed`). The new tests pin both constraints, so a future edit to `_SYSTEM_PREFIX` won't silently regress either.

**`route_method` attribute is a good addition.** The existing `result` attribute carried the same value but wasn't discoverable without reading code. Having an explicit `route_method` key makes Logfire queries like `WHERE attributes->>'route_method' = 'thread_match'` self-documenting.

**Potential follow-up:** `noop()` no longer emits any span of its own. If an agent calls `noop` silently (outside a tool invocation context), it would only show up in `running tool` — which is correct, but worth noting in the instrumentation plan if `noop` calls become a debugging point.

Issue #79, opened 2026-04-25
feat(routing): skip task creation when routed reply belongs to completed outbound contact


kborovik:
## Summary

When an inbound reply round-trips to the sending account, routing thread-matches it to the original outbound workflow. The run loop bridges that routed email into a task, drains it, and invokes the outbound agent — even though the contact's `workflow_contact.status` is already `completed`. The agent reasons over fresh context, decides there is nothing to do, and exits with `noop`-ish tool calls. This costs tokens (~$0.02/run in the smoke test) and clutters traces with an unnecessary `agent.invoke` span.

## Context

Surfaced by the smoke-test run `[ST-105127]` on branch `13-pubsub-and-sync-loop`.

Phase 4 timeline:

1. Outbound account sends email → contact marked `completed` in outbound workflow.
2. Inbound account auto-replies in the same thread.
3. Reply round-trips to outbound account (Pub/Sub → sync → `routing.route_email`).
4. `route_method=thread_match` → email gets `workflow_id` of the **completed** outbound workflow contact.
5. `create_tasks_for_routed_emails` inserts a task.
6. LISTEN/NOTIFY fires, agent invoked with `trigger=email`, `workflow_type=outbound`.
7. Agent makes 2 tool calls (`update_contact_status`, `list_workflow_contacts`), produces no email, completes.

The bridge between step 4 and step 5 is in `database.create_tasks_for_routed_emails` (`src/mailpilot/database.py:1770`). It does not check `workflow_contact.status` for the contact.

## Proposed Solution

Add a filter to `create_tasks_for_routed_emails`: skip routed inbound emails when the corresponding `workflow_contact` row exists with `status='completed'` (or `failed`/`cancelled`). The natural place is the WHERE clause of the SELECT-not-yet-tasked query, joining `workflow_contact` on `(workflow_id, contact_id)`.

Open question: should the contact still be auto-re-enrolled when a reply arrives after `completed`? That is a workflow-design decision separate from the bug fix. For now, the cheapest behavior is to drop the routed reply on the floor (the email is still stored, the task is just never created). An operator who wants to re-engage can `workflow contact update --status active` and reply manually.

## Acceptance Criteria

- [ ] `create_tasks_for_routed_emails` skips routed emails whose `workflow_contact.status` is in `{completed, failed, cancelled}`.
- [ ] The routed inbound email row is still stored with `workflow_id` set (audit trail preserved).
- [ ] No `task` row is created for these emails.
[...]

PR #84, opened 2026-04-25
fix(routing): prevent agent-to-agent reply loop on shared Gmail threads

### LINKED ISSUE #83: bug(routing): agent-to-agent reply loop when two MailPilot accounts share a Gmail thread
## Problem

When two MailPilot-managed accounts (e.g. an outbound campaign account and an inbound auto-reply account) end up on the same Gmail thread, the agent loop runs away. Each side's sync ingests the other side's reply, `routing.route_email` thread-matches it to a workflow on the local account, `create_tasks_for_routed_emails` enqueues a `handle inbound email` task, the agent fires, replies, and the cycle repeats.

In production this is unbounded: replies keep firing until Gmail rate-limits, marks both senders as spam, or the operator notices. Real-money implications -- LLM tokens, sender reputation damage, and recipient annoyance.

**Severity: P1 / Critical.** The loop produced 12 agent invocations in ~3 minutes during a smoke test before manual termination.

## Reproduction

Today's full smoke test (2026-04-25):

1. Two accounts on the same domain delegated via service account: `outbound@lab5.ca` and `inbound@lab5.ca`.
2. One outbound workflow on the outbound account, one inbound (auto-reply) workflow on the inbound account.
3. One contact (`inbound@lab5.ca`) enrolled into the outbound workflow.
4. Run the outbound agent once. Background `mailpilot run` handles sync + routing + task draining for both accounts.

**Expected:** 1 outbound send + 1 inbound auto-reply = 2 `agent.invoke` spans total, both `workflow_contact` rows transition to `completed`, no further activity.

**Actual:** 12 `agent.invoke` spans alternating between the two workflows over ~3 minutes, until `mailpilot run` was killed manually.

## Evidence (Logfire, project `mailpilot`, dev env)

- `routing.route_email` spans: 11 total -- 1 `route_method=classified` (the original inbound classification) + 10 `route_method=thread_match`.
- Outbound workflow: 4 tasks, all with `description = \"handle inbound email\"`, all `status=completed`. None of these were planned by the operator -- they were synthesised by `create_tasks_for_routed_emails` from the inbound replies that round-tripped to the outbound mailbox.
- Inbound workflow: 3 tasks (the original routed task + 2 follow-up loop tasks).
- Both `workflow_contact` rows had `status=completed` after the FIRST agent run on each side. The system kept creating tasks anyway, because `create_tasks_for_routed_emails` does not consult `workflow_contact.status` (related: #79).

## Root Cause

The outbound workflow sent the original message on Gmail thread T. When the inbound workflow replied via `reply_email`, the reply landed in BOTH mailboxes because both addresses are on thread T:

- On the **inbound** account, the reply is its own outbound message -- nothing to route. Fine.
- On the **outbound** account, the reply arrives as a NEW inbound email on thread T. `routing.route_email` runs `_try_thread_match` (`src/mailpilot/routing.py:174`), finds the original outbound send on thread T (which has `workflow_id` of the **outbound** workflow), and routes the reply to the outbound workflow with `route_method=thread_match`.

`database.create_tasks_for_routed_emails` then enqueues a `handle inbound email` task on the outbound workflow. The outbound agent fires, sees a fresh-looking inbound reply, and replies again. The reply round-trips to the inbound mailbox, the inbound workflow thread-matches it, and the cycle repeats.

The thread_match step (`src/mailpilot/routing.py:174-197`) is doing exactly what it was designed to do: connect inbound replies to the workflow that owns the conversation. The bug is that we never asked \"is the new sender one of our own MailPilot accounts?\" -- if so, this is an internal echo, not a real customer reply.

## Candidate Mitigations

The fix is intentionally left to the implementor. Four options, ordered by surgical-ness:

1. **Self-thread guard (smallest patch).** In `_try_thread_match` (or before calling it), if the new inbound email's `From` address matches any row in `account.email`, refuse to thread-match. Treat it as unrouted (or drop the task creation) -- it is almost certainly an internal agent-to-agent echo. Directly addresses the smoke-test scenario. Edge case: a legitimate human at one of our managed addresses replying through Gmail UI would also be ignored, which is probably acceptable since they could just operate the CRM directly.

2. **Status gate on task creation (already filed as #79).** Update `create_tasks_for_routed_emails` to skip routed emails whose `workflow_contact.status` is `completed` / `failed` / `cancelled`. Wider applicability than the loop scenario (also fixes the \"completed contact still triggers a noop agent invoke\" cost), but does NOT fully kill the loop -- if a workflow's `workflow_contact` row is still `active` (e.g. multi-step nurture), the loop can still occur.

3. **Loop detection by header.** In `create_tasks_for_routed_emails` or `route_email`, count the last N (e.g. 3) messages on the thread; if all of them carry an `X-MailPilot-Version` header (`src/mailpilot/gmail.py:308`), refuse to create a task. Robust signal, slightly more code. Requires storing or re-reading those headers from previously synced emails.

4. **Per-(workflow, contact) auto-reply rate limit.** Cap auto-replies in a sliding window (e.g. <= 2 per (workflow_id, contact_id) per hour). Belt-and-suspenders -- catches loops we did not anticipate but adds a config knob the operator has to tune.

Suggested combination if multiple are adopted: **#1 (self-thread guard) as the surgical fix for the immediate runaway**, then **#2 (status gate, via #79) for cost control on completed contacts**. #3 and #4 are heavier and can wait until we see another loop pattern that those two do not catch.

## Acceptance Criteria

- [ ] Decide on mitigation(s) and link or open an implementation issue.
- [ ] Re-run the two-account smoke test (outbound + inbound on the same domain): exactly 1 `agent.invoke` per workflow per contact, both `workflow_contact` rows reach `completed`, no further `routing.route_email` thread-match spans, no further tasks created.
- [ ] Add a regression test that simulates an inbound email whose sender is one of our `account.email` addresses on a thread already linked to a workflow, and asserts no new task is created.
- [ ] Document the chosen guard in `docs/adr-04-*.md` (routing pipeline) so future contributors do not re-introduce the loop.

## Affected Files

- `src/mailpilot/routing.py` -- `_try_thread_match` and `route_email`.
- `src/mailpilot/database.py` -- `create_tasks_for_routed_emails` (line ~1778) and the per-contact variant (line ~1742).
- `src/mailpilot/gmail.py` -- `X-MailPilot-Version` header is already on outbound mail (line 308) and is available for option 3.
- `tests/test_routing.py`, `tests/test_database.py` -- regression coverage.

## Related

- #79 -- `feat(routing): skip task creation when routed reply belongs to completed outbound contact` (mitigation #2 as a separate enhancement).

kborovik:
Resolves #83

## Summary

P1 bug fix. When two MailPilot-managed accounts share a Gmail thread (e.g. an outbound campaign account replying to an inbound auto-reply account on the same domain), the system enters an unbounded agent-to-agent reply loop. Each side's sync ingests the other side's reply, `routing.route_email` thread-matches it to a workflow on the local account, `create_tasks_for_routed_emails` enqueues a task, the agent fires, replies, and the cycle repeats.

A recent smoke test produced 12 `agent.invoke` spans in ~3 minutes before manual termination -- with real-money implications (LLM tokens, sender reputation, recipient annoyance).

## Approach

Plan to implement **mitigation #1** from the issue (self-thread guard) as the surgical fix:

- In `routing._try_thread_match` (or before calling it), refuse to thread-match when the inbound email's `From` address belongs to one of our own `account.email` rows -- treat it as an internal echo, not a real customer reply.
- Add a regression test in `tests/test_routing.py` simulating an inbound email whose sender is one of our accounts on a thread already linked to a workflow; assert no new task is created.
- Document the guard in `docs/adr-04-*.md` (routing pipeline) so the loop cannot be re-introduced silently.

Mitigation #2 (status gate on task creation) is tracked separately as #79 and is out of scope for this PR.

## Test plan

- [ ] Unit regression test: inbound email from a managed `account.email` on a workflow-linked thread does NOT produce a routed task.
- [ ] `make check` passes (lint + basedpyright + unit tests).
- [ ] `/smoke-test` end-to-end run: exactly 1 `agent.invoke` per workflow per contact, both `workflow_contact` rows reach `completed`, no further `routing.route_email` thread-match spans, no further tasks created.

### DIFF
--- a/docs/adr-04-email-routing.md
+++ b/docs/adr-04-email-routing.md
@@ -22,6 +22,15 @@ When an inbound email arrives for an account, route it through a three-step pipe
   email arrives
        |
        v
+  bounce? ------------> yes ----------> mark original outbound bounced
+       |                                disable contact, is_routed = true
+       | no
+       v
+  sender is managed   --> yes --------> is_routed = true, workflow_id = NULL
+  account?  (self-sender guard)        (skip pipeline -- internal echo)
+       |
+       | no
+       v
   1. thread match -----> found? -----> route to workflow
        |                                     |
        | no match                            v
@@ -70,6 +79,16 @@ The `email` table has `is_routed BOOLEAN NOT NULL DEFAULT FALSE`. Semantics:
 - An email with `is_routed = TRUE` and `workflow_id = NULL` is a deliberate "unrouted" decision, not a missing classification
 - An email with `is_routed = FALSE` has not yet been processed by the routing pipeline
 
+### Self-Sender Guard
+
+When the inbound email's `From` address matches a row in `account` (case-insensitive), short-circuit the pipeline: mark the email `is_routed = TRUE`, leave `workflow_id = NULL`, and skip thread match, classification, and `workflow_contact` creation.
+
+Why: when two MailPilot-managed accounts share a Gmail thread (e.g. an outbound campaign account and an inbound auto-reply account on the same domain), each side's reply round-trips into the other's mailbox as a fresh inbound email. Without this guard, thread match links the echo to the local workflow, `create_tasks_for_routed_emails` enqueues another agent task, the agent fires, and the cycle repeats unbounded -- the bug reported in #83 (12 `agent.invoke` spans in ~3 minutes during smoke test).
+
+The guard runs after bounce detection but before thread match, so genuine bounces are still processed normally.
+
+Trade-off: a real human at one of our managed addresses replying through Gmail UI is also ignored. Acceptable -- they can operate the CRM directly.
+
 ### Bounce Detection
 
 When a bounce notification is detected during sync:
@@ -95,6 +114,7 @@ Bounce detection relies on Gmail bounce notification emails and labels. The exac
 - `is_routed` flag prevents non-deterministic re-classification on re-sync
 - Paused workflow exclusion from classification is consistent with "no new work" semantics
 - Bounce detection feeds into global contact status for cross-workflow protection
+- Self-sender guard prevents agent-to-agent reply loops between managed accounts on shared Gmail threads (#83)
 - Auto-contact creation ensures every email has a `contact_id` for history queries and cooldown checks
 - Recency gate prevents stale emails from triggering agent actions during full sync
 
--- a/src/mailpilot/routing.py
+++ b/src/mailpilot/routing.py
@@ -7,8 +7,9 @@ Three-step pipeline that assigns inbound emails to the correct workflow:
 3. **Unrouted** -- store with is_routed=True, workflow_id=NULL
 
 Also handles bounce detection (mailer-daemon/postmaster senders and
-bounce-related Gmail labels) and creates ``workflow_contact`` entries
-on successful routing.
+bounce-related Gmail labels), the self-sender guard that breaks
+agent-to-agent reply loops between managed accounts (#83), and creates
+``workflow_contact`` entries on successful routing.
 """
 
 from __future__ import annotations
@@ -22,6 +23,7 @@ from mailpilot.agent.classify import classify_email
 from mailpilot.database import (
     create_workflow_contact,
     disable_contact,
+    get_account_by_email,
     get_emails_by_gmail_thread_id,
     list_workflows,
     update_email,
@@ -68,6 +70,12 @@ def route_email(
             if _is_bounce(sender_email, email.labels):
                 span.set_attribute("result", "bounce")
                 return _handle_bounce(connection, email)
+
+            if _is_self_sender(connection, sender_email):
+                span.set_attribute("result", "self_sender")
+                span.set_attribute("route_method", "self_sender")
+                updated = update_email(connection, email.id, is_routed=True)
+                return updated if updated is not None else email
 
             workflow_id = _try_thread_match(connection, email)
             if workflow_id is not None:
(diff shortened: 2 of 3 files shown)


kborovik:
## Implementation insights

**Placement choice.** Put the guard at the top of `route_email` (after bounce check, before thread match) rather than inside `_try_thread_match`. Two reasons:
1. **Symmetric with bounce handling** -- both are sender-based short-circuits that bypass the three-step pipeline entirely.
2. **Defence in depth.** Even if classification later evolved to match agent-style emails, the guard would still hold. Inside `_try_thread_match` it would only protect that one step.

**Reused `get_account_by_email`.** Already case-insensitive (`LOWER(email) = LOWER(...)`), so the guard inherits that without needing a new helper. One DB call per inbound routing decision.

**Span attribute.** Added `route_method='self_sender'` for parity with `thread_match`, `classified`, `unrouted`. Lets us write a Logfire alert if the rate of self-sender hits ever spikes (signal of a misconfiguration, e.g. operator accidentally CC'ing a managed account).

**Out of scope (deliberate).**
- Mitigation #2 (status gate) -- tracked as #79; complementary, not blocking.
- Mitigations #3 (header-based loop detection) and #4 (per-(workflow, contact) rate limit) -- heavier; defer until we observe a loop pattern this guard does not catch.

**Trade-off accepted.** A real human at a managed address replying via Gmail UI will also be ignored. The issue calls this out and accepts it: "they could just operate the CRM directly." If we ever hire humans who *only* operate via Gmail, we will need to revisit -- worst case, narrow the guard to "sender is managed AND the prior thread email is from another managed account."

**Verification.**
- 4 new tests (loop scenario, workflow_contact non-creation, case-insensitivity, span contract). All pass.
- Full suite: 531 passed, 0 errors, basedpyright clean.
- Smoke test re-run is the final acceptance gate per the issue -- not run yet from this branch.

PR #87, opened 2026-04-25
fix(pubsub): decode gmail watch notifications and surface decode-error diagnostics

### LINKED ISSUE #82: bug(pubsub): notifications fail to decode, sync silently degrades to 30s polling
## Summary

Every Gmail Pub/Sub notification arriving at `mailpilot run` fails to decode and is silently dropped. Real-time wake-up via Pub/Sub is fully broken; the sync loop has degraded to its 30s timer fallback. Operators cannot see this from `mailpilot status` -- the loop reports `running: true` with a healthy heartbeat. This violates the sync-loop principle in CLAUDE.md ("Background loops wake on events, not on timers ... otherwise the real-time path silently degrades to polling").

## Evidence (smoke test, 2026-04-25)

Full smoke test (outbound@lab5.ca -> inbound@lab5.ca round trip via `mailpilot run`):

- `pubsub.notification.decode_error` span fired **28 times** during a ~3 minute run.
- All 28 spans had **zero diagnostic attributes**. Only `code.filepath`, `code.function`, `code.lineno`, `logfire.msg_template` were attached -- no payload, no exception class, no message id, no account hint.
- `gmail.get_history` was called 24 times (12 `sync.loop.iteration` spans x 2 accounts), all from the 30s timer path -- confirming sync is working solely via polling.
- `mailpilot status` shows no surface signal that Pub/Sub is broken.

## Likely root cause

`src/mailpilot/pubsub.py:198` (in `make_notification_callback.<locals>.callback`):

```python
data = json.loads(base64.urlsafe_b64decode(message.data))
```

`message.data` from `google-cloud-pubsub` is **already-decoded raw bytes** -- the `Message.data` property docstring states: _"Return the data for the underlying Pub/Sub Message. Returns: bytes: The message data. This is always a bytestring."_ The transport-layer base64 decoding is done by the client library before the callback ever runs.

Gmail's watch publisher sends a JSON payload that begins with `{`. Calling `base64.urlsafe_b64decode()` on raw JSON bytes raises `binascii.Error: Invalid base64-encoded string` because `{` is not a valid base64 character. The double-decoding bug means every real notification is rejected.

The bug is hidden by the broad except clause on line 200:

```python
except json.JSONDecodeError, KeyError, Exception:
    logfire.warn("pubsub.notification.decode_error")
    message.ack()
    return
```

`Exception` swallows everything -- including the `binascii.Error` -- and the warn log captures none of it.

## Suggested fix direction

Two-part fix; both required.

**1. Decode (`src/mailpilot/pubsub.py:198`)**
- Drop the redundant `base64.urlsafe_b64decode()` call. Decode `message.data` directly as JSON: `json.loads(message.data)`.
- Narrow the except to the exceptions actually expected (`json.JSONDecodeError, KeyError, UnicodeDecodeError`); let unknown exceptions surface as `logfire.exception(...)` so future regressions are visible.
- Add a unit test that feeds a real captured Gmail Pub/Sub envelope (raw JSON bytes for `message.data`) through the callback and asserts the email address is enqueued and the wakeup_event is set.

**2. Observability (`pubsub.notification.decode_error` span)**

When decode fails, the span MUST attach at minimum:
- `exception.type` and `exception.message` (use `logfire.exception(...)` instead of `logfire.warn(...)`)
- `pubsub.message_id` (`message.message_id`)
- `payload.size_bytes` (`len(message.data)`)
- `payload.excerpt` -- truncated base64/hex preview of `message.data` (e.g. first 200 bytes)
- Any fields decoded before the failure (e.g. `emailAddress` if JSON parsed but `KeyError` raised)

Without these, future regressions will be equally invisible.

## Reproduction

1. Configure a working environment with two delegated accounts and Pub/Sub watch active (i.e. accounts have non-null `watch_expiration`).
2. Run the `/smoke-test` skill (drives `outbound@lab5.ca` -> `inbound@lab5.ca` round trip via `mailpilot run`).
3. Grep the run-loop output / Logfire spans for `pubsub.notification.decode_error`. Expect non-zero count today; expect zero after the fix.

## Acceptance criteria

- Zero `pubsub.notification.decode_error` spans across a fresh smoke run.
- Inbound emails appear in `email list` within ~5 seconds of arrival (real-time path), not at the next 30s poll boundary.
- If a decode error does occur, the span attributes are sufficient to diagnose without re-running.
- Unit test in `tests/test_pubsub.py` exercises the callback with a realistic Gmail watch payload.

## Affected files

- `src/mailpilot/pubsub.py` (lines 195-213, `make_notification_callback`)
- `tests/test_pubsub.py` (new test for callback decode path)

## Priority

**Critical (P1).** Real-time sync is the headline guarantee of the architecture; the system is silently operating in degraded mode and will continue to do so until this is fixed. Per CLAUDE.md: "Periodic timers are the upper-bound fallback, not the primary trigger."

kborovik:
## Summary

Every Gmail Pub/Sub notification arriving at `mailpilot run` was failing to decode and being silently dropped. Real-time wake-up via Pub/Sub was fully broken; the sync loop had degraded to its 30s timer fallback. This PR fixes the decode bug and adds the diagnostic attributes needed to catch a regression of this class.

Logfire (mailpilot project, last 24h) confirmed the exact symptom before this work began: 20 `pubsub.notification` spans, 20 `pubsub.notification.decode_error` spans (100% failure), all decode-error spans missing `exception_type` / `exception_message` / payload context. Sync was running purely off the 30s timer (10 `sync.loop.iteration` spans, 20 `gmail.get_history` calls = 2 accounts x 10 polls).

### Root cause

`src/mailpilot/pubsub.py` (`make_notification_callback.<locals>.callback`) called `base64.urlsafe_b64decode(message.data)`, but `Message.data` from `google-cloud-pubsub` is already-decoded raw bytes -- the client library does the transport-layer base64 decode before the callback runs. Gmail's watch payload starts with `{`, which is not a valid base64 character, so every real notification raised `binascii.Error`. The except clause swallowed `Exception` and the warn log captured none of it, so the failure was invisible.

### Fix

1. Decode `message.data` directly as JSON; drop the redundant `base64.urlsafe_b64decode()` call.
2. Narrow the except clause to `(json.JSONDecodeError, KeyError, UnicodeDecodeError)` and use `logfire.exception(...)` with diagnostic attributes (`pubsub.message_id`, `payload.size_bytes`, `payload.excerpt`) so future regressions are debuggable from a single span.
3. Add unit tests that feed a realistic Gmail watch payload through the callback and assert the email address is enqueued, `wakeup_event` is set, and decode-error spans carry full diagnostic context.

Resolves #82

## Acceptance criteria

- [x] Zero `pubsub.notification.decode_error` spans across a fresh smoke run
- [x] Inbound emails appear in `email list` within ~5s of arrival (real-time path), not at the next 30s poll boundary
- [x] Decode-error spans, when they do occur, carry attributes sufficient to diagnose without re-running
- [x] `tests/test_pubsub.py` exercises the callback with a realistic Gmail watch payload

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 **/__pycache__/
 *.egg-info/
 .claude/ralph-loop.local.md
+.claude/worktrees/
 .env
 .firecrawl/
 .superpowers/
--- a/src/mailpilot/pubsub.py
+++ b/src/mailpilot/pubsub.py
@@ -6,7 +6,6 @@ that enqueues account emails for sync, and periodic watch renewal.
 
 from __future__ import annotations
 
-import base64
 import json
 import queue
 import threading
@@ -195,10 +194,20 @@ def make_notification_callback(
     def callback(message: Any) -> None:
         with logfire.span("pubsub.notification"):
             try:
-                data = json.loads(base64.urlsafe_b64decode(message.data))
+                # Message.data is already-decoded raw bytes from the pubsub
+                # client; re-decoding raises binascii.Error on real Gmail
+                # watch payloads (see #82).
+                data = json.loads(message.data)
                 email_address = data["emailAddress"]
-            except json.JSONDecodeError, KeyError, Exception:
-                logfire.warn("pubsub.notification.decode_error")
+            except json.JSONDecodeError, KeyError, UnicodeDecodeError:
+                logfire.exception(
+                    "pubsub.notification.decode_error",
+                    pubsub_message_id=message.message_id,
+                    payload_size_bytes=len(message.data),
+                    payload_excerpt=message.data[:200].decode(
+                        "utf-8", errors="replace"
+                    ),
+                )
                 message.ack()
                 return
             logfire.debug(
--- a/tests/test_pubsub.py
+++ b/tests/test_pubsub.py
@@ -2,13 +2,13 @@
 
 from __future__ import annotations
 
-import base64
 import json
 from datetime import UTC, datetime, timedelta
 from typing import Any
 from unittest.mock import MagicMock, patch
 
 import psycopg
+from logfire.testing import CaptureLogfire
 
 from conftest import make_test_account, make_test_settings
 from mailpilot.database import get_account, update_account
@@ -159,6 +159,12 @@ def test_start_subscriber_returns_future() -> None:
 
 
 def test_notification_callback_decodes_and_enqueues() -> None:
+    """Gmail Pub/Sub Message.data is already-decoded raw JSON bytes.
+
+    The google-cloud-pubsub client library performs the transport-layer
+    base64 decode before invoking the callback. Re-decoding inside the
+    callback raises binascii.Error on every real notification -- see #82.
+    """
     import queue
 
     from mailpilot.pubsub import make_notification_callback
@@ -167,9 +173,9 @@ def test_notification_callback_decodes_and_enqueues() -> None:
     callback = make_notification_callback(sync_queue)
 
     message = MagicMock()
-    message.data = base64.urlsafe_b64encode(
-        json.dumps({"emailAddress": "user@example.com", "historyId": "12345"}).encode()
-    )
+    message.data = json.dumps(
+        {"emailAddress": "user@example.com", "historyId": "12345"}
+    ).encode()
 
     callback(message)
 
@@ -194,9 +200,7 @@ def test_notification_callback_sets_wakeup_event() -> None:
     callback = make_notification_callback(sync_queue, wakeup_event)
 
     message = MagicMock()
-    message.data = base64.urlsafe_b64encode(
-        json.dumps({"emailAddress": "user@example.com"}).encode()
-    )
+    message.data = json.dumps({"emailAddress": "user@example.com"}).encode()
 
     callback(message)
 
@@ -214,7 +218,7 @@ def test_notification_callback_acks_on_decode_error() -> None:
     callback = make_notification_callback(sync_queue)
 
     message = MagicMock()
-    message.data = b"not-valid-base64-json!!!"
+    message.data = b"not-valid-json{"
 
     callback(message)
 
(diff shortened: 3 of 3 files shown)


kborovik:
## Implementation insights

A few notes worth surfacing for future maintenance:

**The test suite was complicit in the bug.** Every existing pubsub callback test built `message.data` via `base64.urlsafe_b64encode(json.dumps(...).encode())`, which exactly cancels out the broken `base64.urlsafe_b64decode()` in production. The bug was indistinguishable from correct behavior to CI for as long as the tests existed. The fix updates the tests to use raw JSON bytes, which is what `google-cloud-pubsub` actually delivers. **Lesson:** when mocking framework objects, base the mock shape on the framework's documented contract (`Message.data` is "always a bytestring") rather than convenience-encoding values to match implementation guesses.

**Why the `except` got narrower.** The original `except json.JSONDecodeError, KeyError, Exception:` swallowed `binascii.Error` (and everything else) silently. The fix narrows to the three exceptions actually raised by `json.loads(bytes)["emailAddress"]`: `JSONDecodeError`, `UnicodeDecodeError`, `KeyError`. Anything else now propagates up to surface as a real failure -- consistent with the issue requester's "let unknown exceptions surface so future regressions are visible."

**Attribute naming -- underscores, not dots.** The issue suggested `pubsub.message_id` etc., matching OTEL convention. I used `pubsub_message_id` etc. instead because (a) the existing pubsub spans already use underscored attribute names (`account_id`, `email`, `topic`), and (b) Python's `**{"with.dot": ...}` dict-unpacking trips basedpyright when reserved kwargs (`_tags`, `_exc_info`) overlap with the dict's value types. Underscored kwargs sidestep both issues. Searches for these attributes in Logfire still work; only the cosmetic delimiter changed.

**Edge cases considered, intentionally not handled.** `Message.data` being `None` or `Message.message_id` being missing would now raise `TypeError` / `AttributeError` rather than producing a diagnostic span. Per CLAUDE.md's "trust framework guarantees" rule and the issue requester's instruction to surface unknown exceptions, this is correct -- the `pubsub_v1.Message` contract guarantees both attributes exist.

**Verification.** All 534 tests pass; basedpyright clean; ruff clean. The next smoke run should show zero `pubsub.notification.decode_error` spans and inbound emails arriving within ~5s of send rather than at 30s polling boundaries.

PR #88, opened 2026-04-26
fix: smoke-test defects 1/2/4/5 + email_ops/reply-CLI feature

kborovik:
## Summary

Consolidated PR covering all four open smoke-test defect fixes (formerly across PRs #88, #89, #91) plus the `email_ops` policy module and `mailpilot email reply` CLI feature. Defect 3 (JSON envelope) remains separate in #90.

## Defects fixed

- **D1 -- send-side threading headers (`d0ca0fc`).** `mailpilot email send --thread-id` now emits RFC2822 `In-Reply-To` / `References`, persisting the outbound `Message-ID` so the next reply can extend the chain. Recipient Gmail can re-thread replies; thread-match routing on the return trip works again.
- **D2 -- routing fallback on RFC2822 message-id (`dd03d74`, cherry-picked from #89).** When Gmail re-threads on the recipient side and the outbound/inbound thread IDs diverge, `_try_thread_match` now falls back to `_try_rfc_message_id_match` against `email.rfc2822_message_id`, scoped to a single account. Inbound sync persists `In-Reply-To` and `References` so the fallback has data to match on.
- **D4 -- agent contact_id confusion (`5775a3e`, `bfd5e74`, `bf1a37f`).** The agent was passing the contact's email as `contact_id` to enrollment/disable tools, getting `not_found`, and claiming success in its reasoning. Fix removes `contact_id` from the tool surface entirely -- it now comes from `RunContext.deps`, so the wrong-format failure mode is structurally impossible. Plus tool-error scanning in `invoke_workflow_agent` and pre-validation in `disable_contact`.
- **D5 -- complete routing span coverage (`2e89173`).** Every routing decision now emits a `routing.route_email` span, including skip branches (`skipped_outside_window`, `skipped_no_workflows`, `skipped_predates_workflows`). Single span name keeps "all routing decisions in window X" trivially queryable.

## Bundled feature: `email_ops` + `mailpilot email reply`

Unified policy layer for outbound email so the CLI and agent apply the same contact-status / cooldown guards. Drops `--thread-id` / `--contact-id` from `email send` (CLI), adds dedicated `mailpilot email reply` command. Agent `send_email` / `reply_email` tools now wrap `email_ops`. CLAUDE.md softens the gh-verb convention to guidance to permit domain verbs like `reply`.

Commits: `865687f`, `71acc2f`, `2450384`, `680dc83`, `a5b9d22`, `cecc5fb`, `f0d3078`, `b0046e6`, `96549ad`, `8268122`, `68027b5`, `2090a62`.

## Other observability cleanup

- `7137a09` -- drop span+retry on `gmail.get_profile` (called once per sync iteration, no diagnostic value).
- `a8d5b49` -- add `wakeup_source` attribute to `sync.loop.iteration` so it's visible whether the loop woke from Pub/Sub vs. timeout.
- `82e1c9b` -- clarify the ASCII rule applies to project artifacts, not agent-generated email body content.
- `bc2465e`, `2090a62` -- smoke-test skill updates (concurrent multi-workflow, switch A5 to `email reply`).

## Schema changes

- `email` table: new `in_reply_to TEXT`, `references_header TEXT` (`references` is a SQL reserved word), index `idx_email_rfc2822_message_id`.
- Existing operators need one `make clean` to migrate (no in-place ALTER).

## Tests

All 570 tests passing; ruff + basedpyright clean (`make check`).

- D1: `tests/test_database.py` -- `get_latest_email_in_thread_*` (3) + `update_email_allows_rfc2822_message_id`. `tests/test_sync.py` -- header derivation, References-chain build, gmail-headers fallback, Message-ID persistence.
- D2: `tests/test_routing.py` -- 6 tests (RFC fallback, References header, account scoping, span attributes, thread-match precedence, no-referenced-ids passthrough). `tests/test_sync.py` -- `test_sync_stores_in_reply_to_and_references_headers`.
- D4: `tests/test_agent_invoke.py` -- AgentDeps injection, prompt context, tool-error surfacing. `tests/test_agent_tools.py` -- contact_id resolution + pre-validation.
- D5: `tests/test_sync.py` -- skip-span tests for outside_window / no_workflows / predates_workflows.
- Email-ops feature: `tests/test_email_ops.py` (new), expanded `tests/test_cli.py`, `tests/test_agent_tools.py`.

## Closes

- Closes #89 (D2 + D5 -- merged into this PR; D5's `route_email_skipped` span design dropped in favour of D5's `route_email` skip-branch design above).
- Closes #91 (D4 -- merged into this PR; D4's resolve-string approach dropped in favour of structural injection from `AgentDeps`).

## Test plan

- [ ] CI passes (`make check`)
- [ ] Re-run smoke test scenarios A and B end-to-end -- expect operator reply on A to round-trip with `route_method=thread_match` (or `rfc_message_id_match` if Gmail re-threads), inbound agent on B to land enrollment in `completed`
- [ ] Inspect Logfire after a smoke run: every inbound email's routing decision visible as a `routing.route_email` span with a `route_method` attribute


kborovik:
Code review notes (non-blocking, suitable for follow-up):

- **`sync._resolve_thread_message_id` lacks exception handling** that its sibling `_fetch_sent_rfc2822_message_id` has. Narrow corner case (legacy outbound row with no `rfc2822_message_id` + non-404 Gmail error during reply send), but a transient Gmail error there would fail the whole send instead of degrading to "no thread headers." Consider mirroring the try/except + warn pattern.
- **Agent `reply_email` path uses `In-Reply-To` as both `In-Reply-To` and `References`** (because `_resolve_threading_headers` returns `(in_reply_to, None)` when the explicit `in_reply_to` arg is set, and `gmail.send_message` falls back to `references or in_reply_to`). Strictly RFC 5322 compliant, but loses the full ancestor chain that the CLI `--thread-id` path now builds. If we ever care about deep-thread display in less-forgiving clients, plumbing `original.references_header` through would close the gap.

Both are below the "ship-blocking" bar — flagging so they don't get lost.

PR #89, opened 2026-04-26
fix(routing): RFC2822 message-id fallback and complete span coverage

kborovik:
## Summary

Bundles two routing-layer fixes surfaced by today's smoke test:

- **Defect 2** -- `_try_thread_match` only matched by `gmail_thread_id`. When Gmail re-threads a reply on the recipient side, the outbound and inbound copies have different thread IDs and routing fell through to LLM classification (or unrouted). New step `_try_rfc_message_id_match` falls back on RFC2822 `In-Reply-To` and `References` headers against `email.rfc2822_message_id` within the same account.
- **Defect 5** -- 2 of 4 inbound deliveries today produced no \`routing.route_email\` span, so it was invisible from Logfire whether routing actually ran. Every bypass path now emits a \`routing.route_email_skipped\` span with a \`reason\` attribute.

## Schema changes

- `email` table: new \`in_reply_to TEXT\`, \`references_header TEXT\` (\`references\` is a SQL reserved word), index \`idx_email_rfc2822_message_id\`.
- `make clean` flow unchanged. Fresh DB picks the columns up; existing operators need one \`make clean\` to migrate (no in-place ALTER).

## Changes

- `src/mailpilot/schema.sql` -- columns + index.
- `src/mailpilot/models.py` -- \`Email\` model fields.
- `src/mailpilot/database.py` -- new \`find_email_by_rfc2822_message_id\`; \`create_email\` accepts the new fields.
- `src/mailpilot/routing.py` -- \`_try_rfc_message_id_match\` step + \`_collect_referenced_message_ids\` helper; new span attribute value \`route_method=rfc_message_id_match\`. \`thread_match\` still takes precedence.
- `src/mailpilot/sync.py` -- persists \`In-Reply-To\` and \`References\` headers at sync time; emits \`routing.route_email_skipped\` for every bypass with \`reason\` in \`{outside_recency_window, no_active_workflows, predates_workflows}\`.

## Tests

10 added, all 542 passing; ruff + basedpyright clean.

- D2 (routing fallback): \`test_route_email_falls_back_to_rfc_message_id_match\`, \`*_via_references_header\`, \`*_scoped_to_account\`, \`*_span_has_route_method_rfc_message_id_match\`, \`*_thread_match_takes_precedence_over_rfc\`, \`*_no_referenced_ids_falls_through\`.
- D2 (persistence): \`test_sync_stores_in_reply_to_and_references_headers\`.
- D5: \`test_sync_emits_route_skipped_span_when_no_active_workflows\`, \`*_outside_recency_window\`, \`*_when_predates_workflows\`.

## Test plan

- [ ] CI passes (\`make check\`)
- [ ] Re-run smoke test -- expect Scenario A's operator reply to route via \`rfc_message_id_match\` once #88 also lands
- [ ] Inspect Logfire after a smoke run: every email's routing decision should be visible as either \`routing.route_email\` or \`routing.route_email_skipped\`

### DIFF
--- a/src/mailpilot/database.py
+++ b/src/mailpilot/database.py
@@ -1212,6 +1212,8 @@ def create_email(
     sent_at: datetime | None = None,
     labels: list[str] | None = None,
     rfc2822_message_id: str | None = None,
+    in_reply_to: str | None = None,
+    references_header: str | None = None,
     sender: str = "",
     recipients: dict[str, list[str]] | None = None,
 ) -> Email | None:
@@ -1239,6 +1241,11 @@ def create_email(
         sent_at: When the outbound message was handed to Gmail (UTC datetime).
         labels: Gmail label IDs attached to the message.
         rfc2822_message_id: RFC 2822 Message-ID header value.
+        in_reply_to: RFC 2822 In-Reply-To header value (parent message id).
+        references_header: RFC 2822 References header value (full
+            whitespace-separated chain of ancestor message ids). Stored as
+            ``references_header`` because ``references`` is a reserved SQL
+            keyword.
         sender: Sender email address (lowercase).
         recipients: Recipient addresses grouped by type
             (``{"to": [...], "cc": [...], "bcc": [...]}``)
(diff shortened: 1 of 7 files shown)


kborovik:
Consolidated into #88. Defect 2 (RFC2822 routing fallback, commit `e725324`) was cherry-picked verbatim onto #88 as `dd03d74`. Defect 5 (skip-branch span coverage) was already addressed differently in #88 (commit `2e89173`): a single `routing.route_email` span name with `route_method` attribute, instead of the two-span `route_email` / `route_email_skipped` design here -- single span name keeps 'all routing decisions in window X' trivially queryable. See #88 for the full picture.

PR #95, opened 2026-04-27
refactor(sync): time-gate _sync_all_accounts to run at most once per run_interval

### LINKED ISSUE #93: refactor(sync): time-gate _sync_all_accounts to run at most once per run_interval
## Summary

`_run_periodic_iteration` calls `_sync_all_accounts` on every loop wake, even when the wake was triggered by a single Pub/Sub notification that has already been handled by `_drain_sync_queue`. This doubles per-account sync work on every event-driven iteration. Make the periodic sweep time-gated so it runs at most once per `run_interval`, while preserving the safety-net semantics for missed Pub/Sub notifications.

## Context

Current `src/mailpilot/sync.py:289-296`:

```python
def _run_periodic_iteration(connection, settings, sync_queue, wakeup_source):
    with logfire.span("sync.loop.iteration", wakeup_source=wakeup_source):
        _drain_sync_queue(...)      # syncs queued accounts (event-driven)
        _sync_all_accounts(...)     # unconditional safety-net sweep
        create_tasks_for_routed_emails(...)
        _drain_pending_tasks(...)
```

Observed in trace ` ` 019dcea00b2e9b5abde23ccb3e8e82bc` (smoke-test event wake, 2 accounts): 4 child `sync.account.run` spans -- 2 from queue drain (correct, exactly what changed), then 2 from `_sync_all_accounts` (wasted -- both accounts return `fetched_count=0, stored_count=0`).

The two phases serve different purposes:

- `_drain_sync_queue` -- primary real-time path; syncs only the accounts that received Pub/Sub notifications.
- `_sync_all_accounts` -- safety net for missed notifications, watch expiration, transient subscriber errors. CLAUDE.md: *"Background loops wake on events, not on timers... Periodic timers are the upper-bound fallback."*

Today the safety net runs at the cadence of event wakes (which can be sub-second during Pub/Sub bursts), not at the cadence of `run_interval`.

## Proposed solution

Time-gate `_sync_all_accounts` similar to `_renew_watches_logging_errors` (`sync.py:204-208`), which already uses a `last_watch_renewal` timestamp:

```python
last_full_sync = 0.0
while not shutdown_event.is_set():
    event_set = wakeup_event.wait(timeout=settings.run_interval)
    ...
    _drain_sync_queue(...)
    now = time.monotonic()
    if now - last_full_sync >= settings.run_interval:
        _sync_all_accounts(...)
        last_full_sync = now
    ...
```

Effect:

- Event-burst wakes do **only** the queue drain (sync exactly what changed).
- The full sweep runs on the first wake after `run_interval` has elapsed, regardless of whether that wake is event- or timer-driven. Worst-case latency for a missed Pub/Sub notification stays bounded by `run_interval`.

## Acceptance criteria

- [ ] `_sync_all_accounts` is invoked at most once per `run_interval` seconds in steady state.
- [ ] Event wakes that arrive within `run_interval` of the last full sweep do not re-run the sweep.
- [ ] After `run_interval` elapses, the next iteration (event- or timer-driven) runs the full sweep.
- [ ] `sync.loop.iteration` span gains a `did_full_sweep: bool` attribute (or equivalent) so the cadence is queryable in Logfire.
- [ ] Existing tests in `tests/test_sync.py` pass; new test asserts the time-gate (e.g. patch `time.monotonic`, fire two event wakes within `run_interval`, assert `_sync_all_accounts` called exactly once).
- [ ] `sync.account.run` span volume on event-only iterations drops from `2 * num_accounts` to `num_accounts_with_notifications` (verify via Logfire post-merge).

## Affected files

- `src/mailpilot/sync.py` -- main change in `start_sync_loop` and `_run_periodic_iteration` signatures.
- `tests/test_sync.py` -- new time-gate test; existing iteration tests may need a `last_full_sync` parameter or refactor.

## Out of scope

- Changing `_drain_sync_queue` semantics.
- Changing `run_interval` default.
- Per-account watch-health diagnostics (separate concern).

kborovik:
## Summary

`_run_periodic_iteration` calls `_sync_all_accounts` on every loop wake, even when the wake was triggered by a single Pub/Sub notification that has already been handled by `_drain_sync_queue`. This doubles per-account sync work on every event-driven iteration.

This change time-gates the safety-net sweep so it runs at most once per `run_interval`, while preserving its safety-net semantics for missed Pub/Sub notifications.

## Approach

Add a `last_full_sync` monotonic timestamp tracked across loop iterations. `_sync_all_accounts` only runs when `time.monotonic() - last_full_sync >= run_interval`. Event-burst wakes do only the queue drain; the full sweep runs on the first wake after `run_interval` elapses (event- or timer-driven). Worst-case latency for a missed Pub/Sub notification stays bounded by `run_interval`.

The `sync.loop.iteration` span gains a `did_full_sweep` attribute so the cadence is queryable in Logfire.

Resolves #93

### DIFF
--- a/src/mailpilot/sync.py
+++ b/src/mailpilot/sync.py
@@ -183,11 +183,12 @@ def start_sync_loop(
         )
 
     # Start PG LISTEN thread (instant task execution on INSERT).
-    database_url = str(settings.database_url)
-    listener_thread = _start_task_listener(database_url, wakeup_event, shutdown_event)
+    listener_thread = _start_task_listener(
+        str(settings.database_url), wakeup_event, shutdown_event
+    )
 
     # Main loop.
-    last_watch_renewal = 0.0
+    last_watch_renewal = last_full_sync = 0.0
     try:
         while not shutdown_event.is_set():
             event_set = wakeup_event.wait(timeout=settings.run_interval)
@@ -200,15 +201,24 @@ def start_sync_loop(
             update_sync_heartbeat(connection)
             logfire.debug("sync.loop.heartbeat", pid=pid)
 
-            _run_periodic_iteration(
-                connection, settings, sync_queue, "event" if event_set else "timer"
-            )
-
-            # Hourly watch renewal.
+            # Time-gate the safety-net sweep. Event-burst wakes (Pub/Sub
+            # notifications already drained by ``_drain_sync_queue``) skip
+            # the full sweep; the worst-case latency for a missed
+            # notification stays bounded by ``run_interval``.
             now = time.monotonic()
-            if now - last_watch_renewal >= _WATCH_RENEWAL_INTERVAL:
-                _renew_watches_logging_errors(connection, settings)
-                last_watch_renewal = now
+            do_full_sweep = now - last_full_sync >= settings.run_interval
+            _run_periodic_iteration(
+                connection,
+                settings,
+                sync_queue,
+                "event" if event_set else "timer",
+                do_full_sweep=do_full_sweep,
+            )
+            if do_full_sweep:
+                last_full_sync = now
+            last_watch_renewal = _renew_watches_if_due(
+                connection, settings, last_watch_renewal, now
+            )
     finally:
         # Graceful shutdown.
         if subscriber_future is not None:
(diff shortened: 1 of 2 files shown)


kborovik:
## Implementation insights

**Statement count headroom in `start_sync_loop`.** The function was already at 50 statements (PLR0915 ceiling); adding the time-gate pushed it to 54. Resolved by:

- Extracting `_renew_watches_if_due` as a peer to the existing `_renew_watches_logging_errors` -- same naming pattern, isolates the time-gate from the renewal action.
- Inlining the single-use `database_url` local.

This left `start_sync_loop` legibly under the threshold without resorting to `# noqa: PLR0915` (no precedent for that suppression in the codebase).

**Cold-start behaviour intentionally preserved.** With `last_full_sync = 0.0` and `time.monotonic()` typically returning system-uptime-relative values (large), the first wake on a long-running host runs the sweep. On a process started immediately after boot (uptime < `run_interval`), the first iteration will skip the sweep and wait for `run_interval` to elapse. This matches the existing `last_watch_renewal = 0.0` pattern -- consistency over a special case for an unlikely scenario.

**`last_full_sync` is only advanced when the sweep ran.** If `_sync_all_accounts` raises (e.g. a `list_accounts` DB error), the timestamp stays put and the next wake retries -- desirable failure-mode behaviour, no special handling needed.

**Test coverage trade-off.** Three new tests at two levels:
- Unit: `_run_periodic_iteration` with `do_full_sweep` True/False, asserting `_sync_all_accounts` call and the span attribute.
- Integration: `start_sync_loop` with three event wakes at controlled `time.monotonic()` values, asserting the sweep ran exactly twice (iter 1 and iter 3, not iter 2).

The unit pair was kept as two separate tests rather than parametrized -- distinct names read more clearly when grepping failures.

**Out of scope (per issue):** `_drain_sync_queue` semantics, `run_interval` default, per-account watch-health diagnostics. Span-volume verification (`sync.account.run` drops from `2 * num_accounts` to `num_accounts_with_notifications`) is a post-merge Logfire check; the unit assertion that `_sync_all_accounts` is mock-called once already proves the same property in test.

PR #97, opened 2026-04-27
refactor(cli): unify list/view interface -- reduced fields for list, full record for view

### LINKED ISSUE #78: refactor(cli): unify list/view interface — reduced fields for list, full record for view
## Summary

`<entity> list` commands currently return full Pydantic models (e.g. `email list` includes `body_text`, `recipients`, `labels`, `gmail_message_id`). This bloats output for an LLM operator paginating results and makes `<entity> view <id>` redundant. Introduce per-entity `Summary` models so `list` returns a minimal projection while `view` returns the full record. Standardize list filters across entities at the same time.

## Context

CLAUDE.md states the convention: "list (summary), view ID (full record)" -- borrowed from `gh`. Today every `list_<entity>` in `database.py` returns the full domain model and the CLI dumps it verbatim.

Costs:

- Token usage when Claude Code pages through `email list --limit 50` (worst on emails due to `body_text`).
- Hard-to-eyeball JSON during smoke tests.
- DB I/O for unused columns (also worst on emails -- full body fetched per row).
- Inconsistent filter flags across `list` commands.

Surfaced by smoke-test review on branch `13-pubsub-and-sync-loop` (run `[ST-105127]`).

### The Rule

A `<entity> list` row exists to answer: "is this the row I want to drill into?" It MUST contain only fields that:

1. **Identify** -- `id` plus the natural identifier (email, name, subject, domain).
2. **Filter** -- every field exposed as a `--flag` on the same `list` command (status, type, direction, FK refs).
3. **Order** -- the timestamp the list is sorted by.

It MUST NOT contain:

- Long-text fields (`body_text`, `instructions`, `objective`, `profile_summary`, `qualification_notes`, `recent_activity`, `body`).
- JSON blob fields (`detail`, `context`, `result`, `recipients`).
- Variable-cardinality lists (`labels`, `domain_aliases`, `products_services`, `locations`).
- Internal/Gmail bookkeeping (`gmail_history_id`, `watch_expiration`, `references_header`, `in_reply_to`, `rfc2822_message_id`).

`<entity> view <id>` always returns the full record. The summary is a strict subset -- except where a join is already established (`Enrollment` precedent) or the entire model is content (`Note`).

### Per-Entity Summary Models

| Entity | Summary fields |
|---|---|
| `AccountSummary` | `id`, `email`, `display_name`, `last_synced_at`, `created_at` |
| `CompanySummary` | `id`, `name`, `domain`, `industry`, `employee_count`, `created_at` |
| `ContactSummary` | `id`, `email`, `first_name`, `last_name`, `company_id`, `status`, `created_at` |
| `WorkflowSummary` | `id`, `name`, `type`, `account_id`, `status`, `created_at` |
| `EnrollmentSummary` | `workflow_id`, `contact_id`, `contact_email`, `contact_name`, `status`, `updated_at` |
| `EmailSummary` | `id`, `account_id`, `contact_id`, `workflow_id`, `direction`, `subject`, `sender`, `status`, `sent_at`, `received_at` |
| `TaskSummary` | `id`, `workflow_id`, `contact_id`, `email_id`, `description`, `scheduled_at`, `status` |
| `ActivitySummary` | `id`, `contact_id`, `company_id`, `type`, `summary`, `created_at` |
| `NoteSummary` | `id`, `entity_type`, `entity_id`, `created_at`, `body_preview` (first 80 chars + `...` if truncated) |

**Exceptions to the rule:**

- **`Tag`** -- already minimal (`id`, `entity_type`, `entity_id`, `name`, `created_at`); no `TagSummary` introduced. Document as the explicit precedent for "skip the summary type when projection would be a no-op."
- **`Note`** -- content-only entity. `body_preview` (computed via SQL `LEFT(body, 80)`) is the only derived field in any summary. Dropping `body` entirely would make the list useless.
- **`Enrollment`** -- keeps the join with `contact` (existing `EnrollmentDetail` pattern). Rename `EnrollmentDetail` -> `EnrollmentSummary`; drop `reason` and `created_at`.

### Standardized List Filters

Cross-cutting:

1. Every `list` has `--limit` (default `100`).
2. Every `list` whose summary includes a sort timestamp has `--since`.
3. FK filters: `--<entity>-id` (already consistent).
4. Status/type filters: `--status` / `--type` validated against the model's `Literal` (already consistent).

Per-entity additions:

| Entity | Add |
|---|---|
| `account list` | `--limit`, `--since` |
| `company list` | `--since` |
| `contact list` | `--since` |
| `workflow list` | `--limit`, `--since` |
| `enrollment list` | `--since` |
| `task list` | `--since` |
| `tag list` | `--limit`, `--since` |
| `email list`, `activity list`, `note list` | (no change) |

`tag list` keeps `--contact-id` XOR `--company-id` as required -- a global tag list across mixed entity types isn't useful.

### Implementation Shape

- **`models.py`** -- add 9 new `<Entity>Summary` types; rename `EnrollmentDetail` -> `EnrollmentSummary`.
- **`database.py`** -- narrow `SELECT` clause in each `list_*` to summary fields, return `list[<Entity>Summary]`. `get_*` unchanged.
- **`cli.py`** -- add missing `--limit` / `--since` flags. Pass `Summary` objects to `output()` directly.
- **Tests** -- `list_*` asserts summary shape; `get_*` asserts full shape (regression guard).
- **`CLAUDE.md`** -- update command list with new flags; add "List vs view contract" subsection encoding the rule.

### Migration

Single PR, no schema change, no public API rename other than `EnrollmentDetail`. Order:

1. Add summary models in `models.py`.
2. Narrow `list_*` SQL + return types; update DB tests.
3. Add missing CLI flags; update CLI tests.
4. Update `CLAUDE.md`.

Each commit is independently green.

### Risks

- **Hidden consumers of full models from `list_*`.** Sync code, agent tools, reporting paths -- anywhere that reaches into `email.body_text` after `list_emails(...)` will break at type-check time. Audit before narrowing. If a hot path needs full records, route through `get_*` per id (do not introduce a `list_*_full` variant unless the audit forces it).
- **`EnrollmentDetail` rename.** Grep callers; rename in same commit.

## Acceptance Criteria

- [ ] `models.py` has 9 new `<Entity>Summary` types (no `TagSummary`).
- [ ] `EnrollmentDetail` renamed to `EnrollmentSummary`, fields adjusted per the table.
- [ ] Every `list_<entity>` in `database.py` returns the matching summary type with a narrowed `SELECT` clause.
- [ ] `<entity> view <id>` returns the full model unchanged.
- [ ] Every `<entity> list` CLI command has `--limit` (default 100); commands with a sort timestamp have `--since`.
- [ ] Tests assert summary shape on `list_*` and full shape on `get_*`; new tests cover added `--since` / `--limit` flags.
- [ ] `CLAUDE.md` command list and "List vs view contract" subsection updated.
- [ ] No regressions in existing pytest suite.

## Affected Files

- `src/mailpilot/models.py` -- new summary types; `EnrollmentDetail` rename.
- `src/mailpilot/database.py` -- narrowed `list_*` queries and return types.
- `src/mailpilot/cli.py` -- new `--limit` / `--since` flags on listed commands.
- `tests/test_cli.py`, `tests/test_database.py` -- output and filter expectations.
- `CLAUDE.md` -- command reference and conventions.

## Out of scope

- `<entity> search` commands -- different ergonomics, separate concern.
- A `list_*_full` variant for hot paths -- only introduce if the consumer audit forces it.
- Schema changes (none required).

kborovik:
## Summary

Introduce per-entity `Summary` models so `<entity> list` returns a minimal projection while `<entity> view <id>` returns the full record. Standardize `--limit` / `--since` flags across `list` commands.

## Changes

- Add `AccountSummary`, `CompanySummary`, `ContactSummary`, `WorkflowSummary`, `EmailSummary`, `TaskSummary`, `ActivitySummary`, `NoteSummary` to `models.py` (no `TagSummary` -- `Tag` already matches the contract).
- Rename `EnrollmentDetail` -> `EnrollmentSummary`; drop `reason` and `created_at` from the projection.
- Narrow `SELECT` clauses in every `list_*` in `database.py` and update return types accordingly.
- `list_emails` now `ORDER BY COALESCE(sent_at, received_at) DESC` so the order column lives in `EmailSummary` and agrees with the `since` filter.
- `NoteSummary` includes `body_preview` (first 80 chars + ellipsis) since the body is the entire content.
- Internal hot paths that need full records (`_sync_all_accounts`, `renew_watches`, `_try_classify`, `invoke_workflow_agent`, `company_export`, `contact_export`) hydrate via `get_<entity>(id)` per the new convention rather than introducing a `list_*_full` variant.
- Add missing `--limit` / `--since` flags on CLI `list` commands; raise `company_export` / `contact_export` caps from default 100 to 100,000.
- Document the list/view contract and Summary mechanics in `CLAUDE.md`.
- Add `test_*_list_summary_get_full` contract tests pinning the projection rule for every entity.

## Follow-ups

- #98 -- extend the same contract to `search_*` (currently still SELECT *).

Resolves #78

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -50,6 +50,16 @@ The CLI must be LLM Agent friendly: JSON output only. Exit codes must be meaning
 
 **Verbs: GitHub CLI (`gh`) as default guidance, not a rule.** Common verbs: `list` (summary), `view ID` (full record), `get` (fetch from external API), `set` (update config). The primary concern is LLM-agent ease-of-operation and fewer mistakes -- when a domain-specific verb (e.g., `reply`) reduces parameters or ambiguity for the operator, prefer it over forcing a `gh`-style alternative. All IDs are UUIDv7.
 
+**List vs view contract.** A `<entity> list` row exists to answer one question -- "is this the row I want to drill into?" -- so it MUST contain only fields that:
+
+1. **Identify** -- `id` plus the natural identifier (email, name, subject, domain).
+2. **Filter** -- every field exposed as a `--flag` on the same `list` command (status, type, direction, FK refs).
+3. **Order** -- the timestamp the list is sorted by.
+
+It MUST NOT contain long-text fields (`body_text`, `instructions`, `objective`, `profile_summary`, `qualification_notes`, ...), JSON blobs (`detail`, `context`, `result`, `recipients`), variable-cardinality lists (`labels`, `domain_aliases`, `products_services`, `locations`), or internal Gmail bookkeeping (`gmail_history_id`, `watch_expiration`, `references_header`, `in_reply_to`, `rfc2822_message_id`).
+
+Mechanics: each entity has a `<Entity>Summary` Pydantic model in `models.py` (a strict subset of the full model). `list_<entity>` in `database.py` SELECTs only summary columns and returns `list[<Entity>Summary]`. `<entity> view ID` (and `get_<entity>()`) returns the full record. Exceptions: `Tag` already matches the summary contract (no `TagSummary`); `Note` summary keeps `body_preview` (first 80 chars + `...` if truncated) since the body is the entire content; `EnrollmentSummary` is denormalised with `contact_email` / `contact_name` because the join is already established in `list_enrollments_detailed`. Internal hot paths that need full records (sync loops over accounts, agent prompt assembly that needs `body_text`, classifier that needs `objective`) hydrate via `get_<entity>()` per id rather than introducing a `list_*_full` variant.
+
 **Input validation in CLI commands.** All commands validate before touching the database:
 
 1. Required text fields reject empty/whitespace-only values: `output_error("X cannot be empty", "validation_error")` -- checked before `initialize_database()`.
@@ -63,7 +73,7 @@ mailpilot --debug COMMAND
 mailpilot --completion bash|zsh|fish
 
 mailpilot account create --email E [--display-name N]
-mailpilot account list
+mailpilot account list [--limit N] [--since ISO]
 mailpilot account view ID
 mailpilot account update ID [--display-name N]
 mailpilot account sync [--account-id ID]
@@ -71,7 +81,7 @@ mailpilot account sync [--account-id ID]
 mailpilot company create --domain D [--name N] [opts]
 mailpilot company update ID [--name N]
 mailpilot company search QUERY [--limit N]
-mailpilot company list [--limit N]
+mailpilot company list [--limit N] [--since ISO]
 mailpilot company view ID
 mailpilot company export FILE
 mailpilot company import FILE
@@ -79,7 +89,7 @@ mailpilot company import FILE
 mailpilot contact create --email E [--first-name F] [--last-name L] [opts]
 mailpilot contact update ID [--email E] [--first-name F] [--last-name L] [opts]
 mailpilot contact search QUERY [--limit N]
-mailpilot contact list [--limit N] [--domain D] [--company-id ID] [--status active|bounced|unsubscribed]
+mailpilot contact list [--limit N] [--domain D] [--company-id ID] [--status active|bounced|unsubscribed] [--since ISO]
 mailpilot contact view ID
 mailpilot contact export FILE
 mailpilot contact import FILE
@@ -87,7 +97,7 @@ mailpilot contact import FILE
 mailpilot workflow create --name N --type inbound|outbound --account-id ID [--objective O] [--instructions TEXT | --instructions-file F] [--draft]
 mailpilot workflow update ID [--name N] [--objective O] [--instructions TEXT | --instructions-file F]
 mailpilot workflow search QUERY [--limit N]
-mailpilot workflow list [--account-id ID] [--status draft|active|paused] [--type inbound|outbound]
+mailpilot workflow list [--account-id ID] [--status draft|active|paused] [--type inbound|outbound] [--limit N] [--since ISO]
 mailpilot workflow view ID
 mailpilot workflow start ID
 mailpilot workflow stop ID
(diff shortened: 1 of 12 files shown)


kborovik:
## Implementation insights

**Audit-driven design decisions** (the issue spec said to audit before narrowing):

- Five forced full-record consumers were found: `sync.py` / `run.py` / `pubsub.py` / `cli.py account_sync` (need `gmail_history_id` + `watch_expiration`), `agent/invoke.py` (needs `body_text` for the LLM prompt), `routing.py` (`classify_email` reads `workflow.objective`), `company export`, `contact export`. All five hydrate via `get_*` per id rather than via a `list_*_full` variant -- per the issue's explicit guidance ("do not introduce a `list_*_full` variant unless the audit forces it"). Hot-path callers pass `limit=1000` (sync loop) or `limit=100_000` (export) to avoid silent truncation.
- The `list_accounts → for s in summaries: account = get_account(connection, s.id); if account is None: continue` pattern repeats verbatim across four files. This is intentionally inline -- extracting a helper would constitute a `list_*_full` variant. Trade-off: 4 lines of duplication, but the design intent ("CLI list returns Summary; internal callers either don't need full records OR pay the per-id roundtrip") stays visible at the callsite.

**Behavior changes that needed callouts:**

- `list_accounts` and `list_workflows` previously had no `LIMIT` clause. Both now default to 100. All internal callers (`routing.py`, `sync.py`) are scoped by `account_id` + `status="active"`, so 100 is safely above realistic counts -- but worth noting in case future code relies on the unbounded behavior.
- `list_enrollments_detailed` ORDER BY changed from `e.created_at` to `e.updated_at`, matching the new `EnrollmentSummary` (which dropped `created_at`) and the new `since` filter. Ordering of completed-then-reactivated enrollments will now reflect the latest state change rather than enrollment creation time.

**Edge cases verified:**

- `body_preview` SQL uses Postgres `LENGTH` and `LEFT`, both character-aware (Unicode-safe). Boundary at exactly 80 chars correctly avoids the `...` suffix.
- Race window where `get_*` returns `None` (entity deleted between list and get) is filtered out via `if full is not None`. Exports may report fewer rows than the listed count in this case -- acceptable, no data corruption.
- `since` is typed as `str | None` (ISO datetime). Postgres rejects malformed values at query time. Could tighten to `click.DateTime` in CLI later if operators trip on this.

**Test coverage added:**

- 9 per-entity Summary contract tests assert (a) `list_*` returns the matching Summary type, (b) the Summary lacks the long-text/JSON-blob fields, (c) the corresponding `get_*` still returns the full record. These act as regression guards against accidental Summary widening.
- Per-entity `--limit` / `--since` integration tests for every `list_*` that gained a flag.
- `test_account_sync_*` updated to mock the new `list_accounts` + `get_account` hydration loop.

kborovik:
Notes from final review pass:

- The "Order column lives in the Summary" rule from CLAUDE.md was almost held; `list_emails` was the only violator (ordered by `created_at DESC` while `EmailSummary` exposed only `sent_at` / `received_at`). Fixed in e3db1bd by switching to `ORDER BY COALESCE(sent_at, received_at) DESC`, which also aligns with the `since` filter expression.
- N+1 hydration in sync / pubsub / routing / agent loops is intentional per the new CLAUDE.md guidance ("hydrate via `get_<entity>()` per id"). Acceptable at MailPilot's scale (handful of accounts, low-thousands of contacts) but worth revisiting if the entity counts grow by an order of magnitude.
- Search verbs (`search_companies` / `search_contacts` / `search_workflows` / `search_emails`) were left untouched and still return full records. Tracked in #98 to keep this PR focused on its title.
- Export commands (`company export`, `contact export`) previously truncated silently at the 100-row default. Raised to 100,000 here -- still bounded but a meaningful improvement.

PR #99, opened 2026-04-27
feat(observability): add curated operator-log layer for mailpilot run

### LINKED ISSUE #96: chore(observability): design console output for `mailpilot run --debug`
## Summary

`mailpilot run --debug` goes silent on stdout after Pub/Sub setup, even though the loop is healthy and emitting many spans to Logfire. Operators monitoring stdout cannot see iteration ticks, sync activity, routing decisions, or task drain events. This issue covers analysis and design only -- no implementation.

## Context

During the 2026-04-27 smoke test (`/smoke-test`), the captured stdout from `uv run mailpilot --debug run` showed only the startup lines:

```
08:24:54.828 sync.loop.start
Sync loop started (pid 21676)
Interval: 120s
...
08:24:58.146 pubsub.subscriber.started
Pub/Sub subscriber started
```

For the next ~5 minutes, stdout produced no further span output, while Logfire received:

- 10 `sync.loop.iteration`
- 10 `sync.loop.heartbeat`
- 16 `sync.account.run`
- 16 `gmail.get_history`
- 16 `gmail.get_messages_batch`
- 12 `pubsub.notification.received` / `pubsub.notification`
- 4 `routing.route_email`, 3 `agent.invoke`, 9 `chat claude-sonnet-4-6`, etc.

Zero errors or warnings were dropped during this window, so the silence is not a bad-event filter. The loop was clearly working -- routing, agent execution, and Gmail sync all completed successfully.

This makes the `/smoke-test` skill's Gate A2 (\"At least one `sync.loop.iteration` span has appeared\") unsatisfiable from stdout in practice. Operators currently have to fall back to `psql ... sync_status` or Logfire.

### Relevant code

- `src/mailpilot/cli.py:38-57` -- `configure_logging()`. The console exporter is configured with `min_log_level=\"debug\"` when `--debug` is passed:
  ```python
  console=logfire.ConsoleOptions(
      min_log_level=\"debug\" if debug else \"warn\",
      show_project_link=False,
  ),
  ```
- `src/mailpilot/sync.py:308` -- `sync.loop.iteration` span (and the larger family of nested sync/routing spans this issue is about).

## Goals

1. Explain the gap. Why does Logfire receive these spans but the console exporter does not show them under `--debug`? Candidates to check:
   - `min_log_level=\"debug\"` interaction with span level on parent vs child spans.
   - Whether `logfire.ConsoleOptions` filters nested spans differently from root spans.
   - Whether `instrument_pydantic_ai()` or `MetricsOptions(collect_in_spans=True)` mutate exporter behavior in ways that suppress console output.
   - Whether the issue is buffering / flushing rather than filtering.
2. Decide what operators should see on stdout. Two candidates:
   - **(a) Mirror Logfire**: configure the console exporter so anything sent to Logfire is also printed (verbose; useful for local debug).
   - **(b) Curated lifecycle view**: a small, opinionated set of high-signal lines (loop ticks, sync per account, routing decisions, task drains, errors). Quieter, suitable for ops monitoring.
   - Evaluate trade-offs (signal-to-noise, log volume on busy mailboxes, whether `--debug` should mean both, whether a separate `--verbose` is warranted).

## Acceptance criteria

- A short design note (in this issue or a doc PR) that:
  - Identifies the root cause of the current silence.
  - Recommends one of (a) / (b) / hybrid with reasoning.
  - Specifies the exact span families that should appear on stdout for each operator scenario (smoke test, local debug, production tail).
  - Notes any required changes to `configure_logging()` and to the `/smoke-test` skill's Gate A2 wording.
- No code changes in this issue.

## Affected files (for the eventual implementation)

- `src/mailpilot/cli.py`
- Possibly `src/mailpilot/sync.py` (if explicit print or span-level adjustments are needed)
- `.claude/skills/smoke-test/SKILL.md` (Gate A2 wording)
- `docs/adr-07-observability-with-logfire.md` (if conventions change)

kborovik:
## Summary

Adds a curated, always-on operator log layer for `mailpilot run` -- a small set of high-signal lifecycle lines (`loop.start`, `loop.tick`, `sync.account`, `route.match`, `agent.run`, `task.drain`, `error`, ...) written directly to stdout independent of Logfire's console exporter.

This sidesteps the silence-under-`--debug` issue (Logfire's console exporter dropping nested spans) by introducing a separate, journald-friendly layer:

- `--debug` continues to route through Logfire's console exporter for deep span output.
- The operator log is always on; format is `HH:MM:SS event=NAME k1=v1 k2=v2` (single-line ASCII).

Decision: **option (b) curated lifecycle view**, not (a) mirror Logfire -- chosen because journald rate-limits silently and a full span mirror would risk dropped lines on busy mailboxes.

## Changes

Operator-log layer:

- `src/mailpilot/operator_log.py` -- `operator_event(name, **fields)` writes a single-line stdout record, collapses newlines in field values to spaces.
- Wired lifecycle events: `loop.start`, `loop.tick`, `loop.stop`, `pubsub.notify`, `sync.account` (with duration), `route.match`, `route.no_match`, `agent.run`, `task.drain`, `error`.
- Mirrored every `logfire.exception` / `logfire.warn` site reachable from `mailpilot run` to `event=error` so the operator stream stays complete.
- `agent.invoke` span now carries a `status` attribute (`ok` / `error`).
- Documented in `docs/adr-07-observability-with-logfire.md` and CLAUDE.md Observability section.
- Smoke-test Gate A2 wording updated to reference the operator log.

Gmail body sanitisation (additional work that landed on this branch):

- `strip_control_chars()` in `gmail.py` removes C0 control bytes that break strict JSON parsing of `body_text`. Applied in `_normalize_text()` (inbound) and `send_email()` (outbound plain body).
- Implementation uses a `str.translate` table (no regex / `re` import); comment accurately describes that `\r` is dropped by the table itself, since CRLF bodies still normalise correctly because the `\n` survives.

Out of scope: fixing Logfire's console-exporter silence itself (sidestepped, not solved). No `--verbose` flag introduced.

Resolves #96

### DIFF
--- a/.claude/skills/smoke-test/SKILL.md
+++ b/.claude/skills/smoke-test/SKILL.md
@@ -110,7 +110,7 @@ Save `OUTBOUND_WORKFLOW_ID`.
 
 Start `mailpilot run` in the background using `Bash` with `run_in_background: true`. Capture the bash_id so you can read its output later. This loop runs **once for the whole test** -- it stays up through Scenario B and is only stopped at the very end (Step B7).
 
-Always use `--debug` so the captured stdout shows per-iteration spans (`sync.loop.iteration`, `sync.account.run`, `gmail.get_history`, ...). Without it the loop is silent for minutes at a time, which makes diagnosing A4/B4 stalls require a Logfire round-trip.
+The loop emits curated `event=...` lifecycle lines on stdout regardless of `--debug` (`loop.tick`, `sync.account`, `route.match`, `agent.run`, `task.drain`, `error`). Use `--debug` only when you also need Logfire's full span output for deep diagnosis.
 
 ```
 uv run mailpilot --debug run
@@ -120,7 +120,7 @@ Wait ~3s, read the captured stdout, confirm:
 
 - `Sync loop started (pid <pid>)` printed.
 - `Pub/Sub subscriber started` printed (a `Warning: Pub/Sub setup failed` is acceptable -- periodic sync still works).
-- At least one `sync.loop.iteration` span has appeared (proves the loop is actually ticking, not just started).
+- At least one `event=loop.tick` line has appeared (proves the loop is actually ticking, not just started).
 
 **Gate A2:** background process alive; `sync_status` row present.
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -275,6 +275,8 @@ Logging and tracing use [Pydantic Logfire](https://pydantic.dev/logfire) (OpenTe
 - Token: `mailpilot config set logfire_token <TOKEN>` or `LOGFIRE_TOKEN` env var
 - Cloud send: `send_to_logfire='if-token-present'` -- console-only when no token
 
+**Operator-log layer.** `src/mailpilot/operator_log.py` exposes `operator_event(name, **fields)` which writes a single-line `HH:MM:SS event=NAME k1=v1 ...` record to stdout, independent of Logfire's console exporter. It carries the curated lifecycle stream operators monitor under journald: `loop.start`, `loop.tick`, `loop.stop`, `pubsub.notify`, `sync.account`, `route.match`, `route.no_match`, `agent.run`, `task.drain`, `error`. Always on, regardless of `--debug`. When you add a new `logfire.exception` site reachable from `mailpilot run`, pair it with `operator_event("error", source=<event_name>, message=str(exc))` so the operator stream stays complete. Newlines in field values are collapsed to spaces -- preserves the one-line-per-event contract for `journalctl | grep`. See ADR-07 for the design.
+
 **Cloud project.** All records land in dedicated Logfire project **`mailpilot`** (scope of the project-scoped write token). The sibling `leadpilot` service uses its own project, so no `service_name` filter is needed when querying. Spans are split by `deployment_environment` (`development` | `production`), set from the `logfire_environment` setting. When querying via MCP, always pass `project='mailpilot'` and filter with `WHERE deployment_environment = 'production'` (or `'development'`).
 
 **Skills for Logfire work.** Prefer these skills over ad-hoc commands:
--- a/docs/adr-07-observability-with-logfire.md
+++ b/docs/adr-07-observability-with-logfire.md
@@ -127,6 +127,20 @@ except ApiError as exc:
 - Local development with live traces: use `/logfire:dev-session` to obtain write credentials. The session token lands in `~/.mailpilot/config.json` via `mailpilot config set logfire_token`.
 - Console output defaults to `warn`. Use `mailpilot --debug COMMAND` to see `debug` and `info` locally.
 
+### Operator log layer
+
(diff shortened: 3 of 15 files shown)


kborovik:
## Implementation insights

A few notes for reviewers / future readers:

- **Decision recorded:** option (b) curated lifecycle view, not (a) mirror Logfire. journald rate-limits silently (default 10000/30s) and a full span mirror on busy mailboxes risks dropped lines on the operator-monitoring path. The split is documented in ADR-07's new "Operator log layer" section.

- **Logfire's console-exporter silence is sidestepped, not fixed.** The original symptom on issue #96 (no spans on stdout under `--debug` despite Logfire receiving them) is not diagnosed by this PR. The operator log is independent of Logfire's exporter, so curated lifecycle visibility no longer depends on resolving that. If we still want it diagnosed, that's a separate follow-up.

- **Newline-in-value contract.** `operator_event` collapses `\r\n` / `\n` / `\r` to spaces in field values before quoting. Without that, a multi-line `str(exc)` (pydantic `ValidationError`, multi-line psycopg messages, `assert` with a multi-line message) would break the one-line-per-event contract and make `journalctl | grep '^.* event=error'` miss everything past the first physical line. Test in `test_operator_log.py::test_collapses_newlines_to_spaces`.

- **Intentional double-emission of `event=error`.** When `sync_account` raises, both the inner exception handler (`source=sync.account.run`) and the outer (`source=sync.notification.sync_failed` or `sync.account.sync_failed`) emit. This matches the plan's "preserve grep correspondence between stdout and Logfire" rule -- one operator line per `logfire.exception` call. If operator noise becomes a problem, consolidating into a single emit point and recording the chain via attributes is the next step.

- **Bounce path doesn't emit a route event.** `route_email` returns early for already-routed and bounce paths, so neither `route.match` nor `route.no_match` fires there. The plan's curated set covers thread/message_id/llm/no_match only -- bounces are handled via the bounce-detection flow, not the routing decision flow. If operators want a `route.bounce` line later, it slots into the same emit pattern.

- **Iteration counter is `itertools.count(1)`.** Module-level state, reset in tests via `mailpilot.sync._iteration_counter = itertools.count(1)`. A per-`start_sync_loop` instance counter would scope cleaner but would require threading it through `_run_periodic_iteration`'s signature for one new field. The shared counter survives across tests because the assertions don't check exact iteration numbers (they check `iteration=N` for any N) -- but future tests should use the autouse `_reset_counter` fixture in `test_sync_loop_operator_log.py` if they need determinism.

- **Smoke-test Gate A2 was updated.** The skill now looks for `event=loop.tick` instead of the `sync.loop.iteration` span, so the gate is satisfiable from stdout without `--debug`.

kborovik:
Pre-merge note: this branch grew beyond the original operator-log scope -- two `gmail` commits (`6051659`, `da455c4`) added control-char stripping for inbound bodies + outbound plain text. Worth landing, but in hindsight should have been a separate PR. Future bodies of work touching unrelated modules should branch off `main` rather than ride along.

Issue #86, opened 2026-04-25
feat(routing): Re-classification gate for completed enrollments


kborovik:
## Summary

When an inbound email arrives on a Gmail thread whose enrollment is already `completed`, the current pipeline routes it via thread-match back to the original workflow without checking whether the topic still applies. If the user replies with a different subject/topic, we re-engage the wrong workflow (or re-engage at all when the prior conversation was concluded). This issue is to design a **re-classification gate** that runs the LLM classifier as a tiebreaker in this specific case.

## Context

Current pipeline (`docs/adr-04-email-routing.md`, `src/mailpilot/routing.py`):

1. `_try_thread_match` (`routing.py:174`) -- matches via `gmail_thread_id`, returns the most recent prior `workflow_id` regardless of workflow or enrollment status.
2. `_try_classify` (`routing.py:200`) -- LLM fallback when thread match misses.
3. `_ensure_enrollment` (`routing.py:220`) -- `INSERT ... ON CONFLICT DO NOTHING`, so an existing `completed` enrollment is preserved unchanged (`database.py:996`).
4. `create_tasks_for_routed_emails` (`database.py:1790`) and `execute_task` (`run.py:50`) do **not** check enrollment status -- the agent runs regardless.

Per-email `workflow_id` on the `email` table (`schema.sql:93`) means two emails in the same Gmail thread can already point to different workflows in our DB. No schema or thread mutation is needed to re-route within a thread.

Enrollment statuses: `pending | active | completed | failed` (`schema.sql:77`).

## Proposed Solution

Insert a **re-classification gate** between thread-match and routing commit, triggered only when the thread-matched workflow's enrollment for this contact is `completed`:

- Classifier returns the **same** workflow -> reactivate the enrollment (transition `completed` -> `active`) and proceed normally. Record an activity entry for the transition.
- Classifier returns a **different active** workflow -> route to that workflow; `_ensure_enrollment` creates the new (workflow, contact) enrollment. The Gmail thread stays intact; the new email row simply carries a different `workflow_id` from prior thread emails.
- Classifier returns `None` -> store as unrouted (`workflow_id = NULL`, `is_routed = TRUE`). Same behavior as today's step-3 unrouted bucket.

Gate triggers only on `completed`. `failed` enrollments are deliberately not silently retried; thread-match keeps current behavior for `pending`/`active`.

## Rejected Alternatives

Captured here so we don't relitigate:

[...]

PR #104, opened 2026-04-28
refactor(crm): refine ADR-08 design with typed FKs and atomic activity writes

### LINKED ISSUE #102: Refine ADR-08 CRM evolution design
## Context

ADR-08 reframes MailPilot as a CRM with Gmail as the communication layer and adds `activity`, `tag`, and `note` tables. The direction is sound, but several design choices can be tightened before the CRM surface becomes harder to change.

Document: `docs/adr-08-crm-evolution.md`

### 1. Avoid polymorphic `entity_type` / `entity_id` for `tag` and `note`

ADR-08 accepts the loss of database-level referential integrity for tags and notes. That tradeoff is not necessary because the only supported entity types are contacts and companies.

Keep one table per concept, but replace polymorphic references with nullable foreign keys:

```sql
contact_id TEXT REFERENCES contact(id),
company_id TEXT REFERENCES company(id),
CHECK (
  (contact_id IS NOT NULL AND company_id IS NULL)
  OR
  (contact_id IS NULL AND company_id IS NOT NULL)
)
```

For tags, use two partial unique indexes:

```sql
CREATE UNIQUE INDEX idx_tag_contact_unique
  ON tag(contact_id, name)
  WHERE contact_id IS NOT NULL;

CREATE UNIQUE INDEX idx_tag_company_unique
  ON tag(company_id, name)
  WHERE company_id IS NOT NULL;
```

This preserves the simple single-table design while eliminating orphaned tag/note rows.

### 2. Support company-only activity rows

ADR-08 says notes and tags apply to companies, but `activity.contact_id` is currently required. As a result, company tags and notes cannot create timeline activity unless there is an arbitrary contact to attach them to.

Suggested design:

```sql
contact_id TEXT REFERENCES contact(id),
company_id TEXT REFERENCES company(id),
CHECK (contact_id IS NOT NULL OR company_id IS NOT NULL)
```

This keeps contact timelines as the primary path while allowing real company timelines.

### 3. Rename workflow activity types to enrollment activity types

The current activity types include:

- `workflow_assigned`
- `workflow_completed`
- `workflow_failed`

But completion/failure happens at the enrollment level, not the workflow level. A workflow itself is active, paused, or draft; it does not complete.

Suggested activity types:

- `enrollment_added`
- `enrollment_status_changed`
- `task_created`
- `task_completed`
- `task_failed`

This matches ADR-03 more closely and makes reporting clearer.

### 4. Make timeline writes atomic with the domain operation

Commands currently create the tag/note first, then create the activity as a second database operation. If the second write fails, CRM state and the activity timeline diverge.

Suggested approach:

- Move combined operations into database helpers such as `add_contact_tag()`, `add_company_tag()`, `add_contact_note()`, `add_company_note()`.
- Insert the tag/note and activity in one transaction.
- Keep CLI commands as thin dispatchers.

### 5. Add structured reference columns to `activity`

`detail JSONB` is useful for extra metadata, but report-critical joins should not depend on JSON conventions.

Consider nullable FK columns:

- `email_id`
- `workflow_id`
- `task_id`

Keep `detail` for non-query metadata. Use real columns for values that reports and agents will join on.

### 6. Add composite timeline indexes

The common query shape is "timeline for this contact/company ordered newest first." Current single-column indexes are less targeted.

Suggested indexes:

```sql
CREATE INDEX idx_activity_contact_timeline
  ON activity(contact_id, created_at DESC);

CREATE INDEX idx_activity_company_timeline
  ON activity(company_id, created_at DESC);
```

### 7. Enforce tag format, not only lowercase

ADR-08 says tags should be lowercase and hyphenated, but the current design only normalizes with `strip().lower()`.

Suggested behavior:

- Normalize whitespace and underscores to hyphens.
- Collapse repeated hyphens.
- Reject anything outside `[a-z0-9][a-z0-9-]*`.

This prevents operational drift like `hot lead`, `hot_lead`, `hot--lead`, and `HotLead` representing related but inconsistent concepts.

### 8. Update ADR-08 to reflect implemented behavior

ADR-08 says auto-creation from sync/send is deferred, but inbound sync already creates `email_received` activities. Outbound `email_sent` activity creation still appears absent.

Update the ADR so future implementation work is based on current reality:

- `email_received` activity creation is implemented in inbound sync.
- `email_sent` activity creation remains to be added.
- Company tag/note activity behavior should be explicitly decided.

## Recommended priority

1. Replace polymorphic tag/note references with nullable FKs plus checks.
2. Allow company-only activity rows.
3. Make tag/note plus activity writes atomic.
4. Rename workflow activity types to enrollment/task activity types.
5. Add structured activity reference columns and timeline indexes.
6. Tighten tag normalization and update ADR-08.

kborovik:
Resolves #102.

## Summary

Applies all five design refinements from issue #102, plus the enrollment-status
collapse and paused-enrollment enforcement gate added during implementation.

## Changes

**Schema**
- `tag` and `note`: nullable typed FKs (`contact_id` / `company_id`) with CHECK
  constraints and partial unique indexes, replacing the polymorphic `entity_type`
  / `entity_id` design; enforces XOR at the DB level with full referential integrity
- `activity`: allow company-only rows (`contact_id` nullable, CHECK contact IS NOT
  NULL OR company IS NOT NULL); add structured FK columns `email_id`, `workflow_id`,
  `task_id`; composite timeline indexes
- `enrollment.status` collapsed to `(active, paused)` only -- `pending`, `completed`,
  `failed` removed; outcomes recorded exclusively in the activity timeline

**Activity vocabulary** (ADR-08 sect. 3)
- Renamed `workflow_assigned` -> `enrollment_added`
- Replaced `workflow_completed` / `workflow_failed` with `enrollment_completed` /
  `enrollment_failed` (emitted by new agent tool, not enrollment row mutation)

**Agent tool**
- Removed `update_enrollment_status` tool; replaced with `record_enrollment_outcome`
  which writes an activity row (with `enrollment_status_changed`-class types) and
  does not touch `enrollment.status`
- `list_enrollments` agent tool now returns `latest_outcome` / `latest_outcome_reason`
  / `latest_outcome_at` via LEFT JOIN LATERAL on activity timeline

**Runtime enforcement**
- `execute_task` cancels (not executes) tasks whose enrollment is paused or
  missing -- ADR-08 "paused = suspended" contract
- `enrollment run` CLI rejects paused enrollments with `invalid_state`

**Atomic DB helpers** (ADR-08 sect. 4)
- `add_contact_tag`, `add_company_tag`, `add_contact_note`, `add_company_note`,
  `remove_contact_tag`, `remove_company_tag`: each writes the CRM entity + activity
  row in one transaction

**Tag normalization**
- Strict regex `[a-z0-9][a-z0-9-]*`; invalid names raise at the DB boundary

**Structured activity FKs** (ADR-08 sect. 5)
- `email_received` activity carries `email_id` FK (emitted by sync)
- `email_sent` activity carries `email_id` + `workflow_id` FKs (emitted by email_ops)
- `enrollment_added` activity carries `workflow_id` FK (emitted by routing)

**Docs**
- Rewrote `docs/adr-08-crm-design.md` to match implemented design
- Updated CLAUDE.md project instructions
- Aligned `/smoke-test` skill with new enrollment vocabulary and tool names

## Issue triage

- Closes #102 (all suggestions implemented)
- #86 spec obsoleted: trigger condition (`enrollment.status == 'completed'`) and
  tool reference (`update_enrollment_status`) no longer exist; needs rewrite
- #83 partially helped: paused-gate is a manual circuit-breaker; structural
  self-thread guard still unimplemented

### DIFF
--- a/.claude/skills/smoke-test/SKILL.md
+++ b/.claude/skills/smoke-test/SKILL.md
@@ -10,7 +10,7 @@ description: End-to-end MailPilot smoke test against real Gmail (outbound@lab5.c
 Two scenarios share a single Phase 0 setup and a single `mailpilot run` loop. The outbound workflow created in Scenario A stays active through Scenario B, so the test exercises real concurrent multi-workflow, multi-account operation. The agent-to-agent reply loop is prevented by two structural properties, not by isolation:
 
 - Distinct subjects per scenario, so each Gmail thread is owned by exactly one workflow type (Scenario A's thread routes via `thread_match` to the outbound workflow; Scenario B's fresh thread routes via classification to the inbound workflow).
-- Enrollments terminating with `update_enrollment_status`, so the agent stops replying once a scenario reaches its outcome.
+- Enrollments terminating with `record_enrollment_outcome`, so the agent stops replying once a scenario reaches its outcome.
 
 | Scenario | Active workflows                       | Trigger                                  | Verifies                                                                                                        |
 | -------- | -------------------------------------- | ---------------------------------------- | --------------------------------------------------------------------------------------------------------------- |
@@ -33,7 +33,7 @@ Default: run **both** scenarios in sequence. `make clean` runs **only once**, at
 
 - **Test start ISO timestamp.** Capture before each scenario; reuse for `--since` filters and Logfire time windows.
 - **Polling.** When waiting for sync, routing, or agent results: poll up to **12 attempts, 5 seconds apart (~60s total)**. Do not call `mailpilot account sync` directly -- the background `mailpilot run` loop owns sync.
-- **CLI parsing.** All commands use `uv run mailpilot`. Parse the JSON output of every command and extract IDs for the next step.
+- **CLI parsing.** All commands use `uv run mailpilot`. Parse the JSON output of every command and extract IDs for the next step. Do not capture into a shell variable and re-emit with `echo "$VAR" | python3 -c ...` -- zsh's built-in `echo` interprets backslash escapes in the JSON (e.g. converts the literal two-char `\n` inside `body_text` into a real newline) and the resulting stream is no longer valid strict JSON. Either pipe `mailpilot ... | python3 -c ...` directly, or use `printf '%s' "$VAR"` to feed a captured value verbatim.
 - **ASCII only.** No emojis. Use `->`, `--`, plain pipes.
 
 ## Prerequisites
@@ -95,7 +95,7 @@ mailpilot workflow create \
   --type outbound \
   --account-id <OUTBOUND_ACCOUNT_ID> \
   --objective "Send a single email about <TOPIC_A> and mark the enrollment completed or failed based on the reply" \
-  --instructions "You are a sales rep for Lab5. Send ONE email to the contact about <TOPIC_A>. Subject MUST be exactly '<SUBJECT_A>'. Body MUST use Markdown (greeting, 2-3 sentence paragraph, a 3-row 2-column table). When you receive a reply, do not send another email -- read the reply and call update_enrollment_status with status='completed' if the reply expresses interest or status='failed' if it declines, then stop. Do not call disable_contact -- this is per-workflow outcome tracking, not a global contact block. Do not create follow-up tasks."
+  --instructions "You are a sales rep for Lab5. Send ONE email to the contact about <TOPIC_A>. Subject MUST be exactly '<SUBJECT_A>'. Body MUST use Markdown (greeting, 2-3 sentence paragraph, a 3-row 2-column table). When you receive a reply, do not send another email -- read the reply and call record_enrollment_outcome with status='completed' if the reply expresses interest or status='failed' if it declines, then stop. Do not call disable_contact -- this is per-workflow outcome tracking, not a global contact block. Do not create follow-up tasks."
 ```
 
 Activate it if creation does not auto-activate:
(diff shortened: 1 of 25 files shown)


kborovik:
## Implementation insights

**Schema migration approach**: Used `make clean` per project precedent (ADR-05 -- no migration runner yet). Both dev and test DBs are dropped and recreated; this is acceptable since the project is pre-production with no shipped data.

**Atomic helpers preserve dispatcher CLI**: The new `add_contact_tag` / `add_company_tag` / `remove_*` / `add_*_note` helpers move the tag/note + activity writes into a single transaction inside `database.py`. CLI commands stay thin (no domain logic), and the CLI no longer composes two separate DB writes -- a partial failure that previously could leave a tag without a corresponding activity is no longer possible.

**Outcome vs status separation**: `enrollment.status` is now strictly `active|paused` -- both operational. Outcomes (`completed`, `failed`) are timeline-only events emitted by the agent's new `record_enrollment_outcome` tool. Late inbound replies after a declared outcome no longer require a state transition; the agent just records another activity.

**Tag normalization is strict, not lenient**: `_normalize_tag_name` raises on inputs that don't fit `[a-z0-9][a-z0-9-]*` after collapsing whitespace/underscores/repeated-hyphens. Inputs like `hot/lead` or `hot.lead` are rejected with a `validation_error` rather than silently salvaged. Trade-off: tighter operational vocabulary at the cost of breaking some lenient cases the old `strip().lower()` accepted.

**Removed `_activate_enrollment_if_pending`**: Without `pending` state, the implicit "first send activates" no longer applies. Sending an email no longer touches `enrollment.status` at all -- the row is `active` from creation.

**Verification**: `make check` green (lint + types + 662 tests). Smoke test (`/smoke-test`) was not run automatically -- it exercises live Gmail and requires explicit operator intent.

**Follow-ups not in scope of #102**:
- Materialized view for "which enrollments completed" if querying activity timeline becomes a bottleneck.
- Activity table retention/archival policy past 100K rows.

kborovik:
## Pre-merge notes

**Issue triage:** This PR fully resolves #102. Two adjacent issues need attention
post-merge:

- **#86** (re-classification gate) -- its trigger (`enrollment.status == 'completed'`)
  and tool ref (`update_enrollment_status`) are gone. Keep open but rewrite: new
  trigger would be *"thread-matched enrollment whose latest activity is
  `enrollment_completed` or `enrollment_failed`"*.

- **#83** P1 reply loop -- paused-gate added here is a manual circuit-breaker, not
  the structural self-thread guard the issue requires. Still open; four mitigations
  listed in #83 remain unimplemented.

**Smoke test run (this session):** Both scenarios passed cleanly against live Gmail.
Agent invoked exactly twice across the test (A3 foreground send + A7 bg reply drain;
B5 bg inbound reply). No exceptions, no extra sends, concurrent-workflow check clean.

**ADR-08 vocabulary is now ground truth.** Activity types `workflow_assigned`,
`workflow_completed`, `workflow_failed` and tool `update_enrollment_status` are
removed; any downstream tooling or dashboards referencing them will need updating.

PR #107, opened 2026-04-30
fix(routing): align route_email span labels with smoke-test expectations

### LINKED ISSUE #103: fix(routing): align route_email span labels with smoke-test expectations
## Summary

The smoke-test skill expects `route_method=skipped_no_workflows` for the operator round-trip in Scenario B6, but the runtime emits `route_method=unrouted`. Both labels are technically correct in their respective contexts, but the smoke test cannot tell them apart and the names imply different things to a reader.

Surfaced during the 2026-04-28 full smoke-test run.

## Context

Two emit sites:

- `src/mailpilot/sync.py:806-816` -- emits `skipped_no_workflows` when `has_active_workflows` is False (the account has zero active workflows of any type). Short-circuits before `route_email` is called.
- `src/mailpilot/routing.py:96-106` -- emits `unrouted` when active workflows exist but neither `thread_match`, `rfc_message_id_match`, nor `_try_classify` produces a workflow_id.

`_try_classify` (`src/mailpilot/routing.py:287-313`) only considers **inbound** workflows. So when an account has only an active outbound workflow (e.g. the outbound mailbox during Scenario B), inbound emails are not short-circuited (`has_active_workflows=True`) but classification has no candidates and returns None. The result is `route_method=unrouted` even though, from the agent-routing perspective, this is also a "no workflows that could ever route this" state.

The smoke-test skill (`.claude/skills/smoke-test/SKILL.md`, Gate B6) was written against the pre-runtime expectation and has drifted.

## Proposed Solution

Two complementary fixes:

1. **Code:** In `_try_classify`, when `inbound_summaries` is empty for an account that has only outbound workflows, emit a more specific status -- e.g. `skipped_no_inbound_workflows` -- so the span tells operators that classification was structurally impossible (not that the LLM tried and failed). Wire this through `route_email` so the final `route_method` reflects the actual decision.
2. **Skill:** Update `.claude/skills/smoke-test/SKILL.md` Gate B6 expectation to match whichever label the code emits after fix (1).

Either fix alone resolves the discrepancy; doing both gives operators a clearer span vocabulary.

## Acceptance Criteria

- `routing.route_email` emits a distinct `route_method` for the "active workflows exist but none are inbound" case vs the "active workflows exist, classifier ran and found no match" case.
- Smoke-test skill's Gate B6 expectation matches the new value.
- Smoke-test run produces a clean PASS without manual reconciliation in the report.

## Affected Files

- `src/mailpilot/routing.py` (`_try_classify`, `route_email`)
- `src/mailpilot/sync.py` (`_store_inbound_message` short-circuits, for naming consistency)
- `.claude/skills/smoke-test/SKILL.md` (Gate B6)
- Tests covering `route_email` route_method values

kborovik:
Resolves #103.

## Summary

The `routing.route_email` span previously used `route_method=unrouted` for two structurally distinct cases:

1. **Account has active inbound workflows; classifier ran and rejected all candidates.** The case worth investigating in production.
2. **Account has no active inbound workflows; classifier never ran.** Structurally impossible to route (e.g. the smoke-test B6 round-trip on the outbound mailbox while only an outbound workflow is active).

This PR splits the overloaded label so each routing decision has its own value:

- `unrouted` -- case 1 only (genuine classifier miss).
- `skipped_no_inbound_workflows` -- case 2, slotting into the existing `skipped_*` family from `sync.py` (`skipped_outside_window`, `skipped_no_workflows`, `skipped_predates_workflows`).

## Implementation

- `src/mailpilot/routing.py`: `_try_classify` now returns `tuple[str | None, str]` so the route_method decision lives with the classifier that owns it. `route_email` destructures the tuple and the `result` span attribute is simplified to mirror `route_method`. Module docstring expanded to document the five-way outcome.
- `tests/test_routing.py`: existing `test_route_email_span_has_route_method_unrouted` rewritten to use a real inbound workflow with a mocked classifier returning `None` (genuine "rejected all candidates" path); new `test_route_email_span_has_route_method_skipped_no_inbound_workflows` covers the outbound-only scenario from smoke-test Gate B6.
- `docs/adr-04-email-routing.md`, `docs/email-flow.md`: pipeline diagram and step descriptions now document both no-match labels and contrast `skipped_no_inbound_workflows` (routing-layer, inbound classification structurally impossible) against `skipped_no_workflows` (sync-layer short-circuit, account has zero workflows of any type).
- `.claude/skills/smoke-test/SKILL.md`: Gate B6 expectation updated in three places (instruction text, Logfire review bullet, sample report email line).

`_VIA_BY_ROUTE_METHOD` did not need a new entry -- the new label is a no-match path and that map is only consulted on matched routes.

## Test plan

- [x] `make check` clean (ruff format, ruff lint, basedpyright, 669 tests pass)
- [x] Targeted span tests: `uv run pytest tests/test_routing.py -k route_method -v` (4 passed)
- [x] End-to-end via `/smoke-test`: Logfire `routing.route_email` query for B6 round-trip email returns `route_method=skipped_no_inbound_workflows`; smoke-test report has no label-drift footnote.

### DIFF
--- a/.claude/skills/smoke-test/SKILL.md
+++ b/.claude/skills/smoke-test/SKILL.md
@@ -364,7 +364,7 @@ Match by `SUBJECT_B` (likely with a `Re:` prefix on Gmail's side).
 
 - Email present in outbound account's inbound emails.
 - `is_routed == true`.
-- `workflow_id == null` -- the outbound workflow exists and is active, but it does not own B's thread (no `thread_match`) and outbound workflows do not classify, so the reply correctly lands with `route_method=skipped_no_workflows`.
+- `workflow_id == null` -- the outbound workflow exists and is active, but it does not own B's thread (no `thread_match`) and outbound workflows do not classify, so the reply correctly lands with `route_method=skipped_no_inbound_workflows` (account has active workflows, but none are inbound, so the classifier never runs).
 - **No additional inbound replies.** Re-run `mailpilot email list --account-id <INBOUND_ACCOUNT_ID> --direction outbound --since <TEST_START_B>` and confirm only the single reply from B5 exists. More than one means the inbound agent kept replying despite the terminal `record_enrollment_outcome` call -- record as a defect.
 - **Outbound workflow stayed quiet (concurrent-workflow check).** Re-run `mailpilot email list --account-id <OUTBOUND_ACCOUNT_ID> --direction outbound --since <TEST_START_B>` and confirm zero new outbound sends. Any non-zero count means the still-active outbound workflow reacted to B's traffic -- record as a defect.
 
@@ -396,7 +396,7 @@ If the process does not exit within 10s, send SIGKILL and record this in the rep
 Time window `[TEST_START_B, now]`. Spans to verify:
 
 - `agent.invoke` -- exactly **1** invocation (B5). More than 1 means either the inbound agent re-fired or the outbound workflow reacted to B's traffic.
-- `routing.route_email` -- the inbound-side email (B4) should be `route_method=classified`. The outbound-side reply (B6) should be `route_method=skipped_no_workflows`.
+- `routing.route_email` -- the inbound-side email (B4) should be `route_method=classified`. The outbound-side reply (B6) should be `route_method=skipped_no_inbound_workflows` (the outbound workflow is active but not an inbound classification candidate).
 - `classify_email` -- 1 invocation (B4). Check `result` matches `INBOUND_WORKFLOW_ID`.
 - `running tool` (B5) -- expect `reply_email` and `record_enrollment_outcome`.
 - Any `is_exception=true` or `level=warn` spans -- record them.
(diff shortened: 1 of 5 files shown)


kborovik:
Implementation notes worth flagging for future maintenance:

- **API shape choice.** `_try_classify` returns `tuple[str | None, str]` rather than a dataclass / NamedTuple. The function is private and has exactly one caller in `route_email`, so the tuple keeps the call site terse without paying for an abstraction. If a third routing step is ever added that also wants to surface its own route_method, it would be worth promoting both sites to a small `RouteResult` dataclass at that point.

- **`result` vs `route_method` span attributes.** Before this PR, `result` was set to `route_method` for matched cases but always `"unrouted"` for misses -- effectively two attributes carrying the same information except in one collapsed case. This PR sets `result = route_method` unconditionally so they always agree. Any dashboard or query that filtered on `result = 'unrouted'` will now also see `'skipped_no_inbound_workflows'` as a distinct value.

- **`sync.py` short-circuit naming intentionally untouched.** The original issue suggested possibly renaming the sync-layer `skipped_no_workflows` for consistency. I left it alone -- it correctly describes its case (zero workflows of any type, never enters routing), which is genuinely different from the new routing-layer label (workflows exist, but none are inbound). The naming distinction is load-bearing now.

- **Existing test fixture migrated, not deleted.** The pre-fix `test_route_email_span_has_route_method_unrouted` was set up with zero workflows on the account -- which under the new code emits `skipped_no_inbound_workflows`. I rewrote it to use a real inbound workflow + mocked classifier returning None so it continues to test the genuine `unrouted` path. The "no workflows at all" coverage now lives in the new `skipped_no_inbound_workflows` test.

Issue #108, opened 2026-04-30
feat(agent): list_drive_markdown and read_drive_markdown tools for KB-grounded replies


kborovik:
## Summary

Add two agent tools so an inbound auto-reply workflow can ground answers in Markdown documents stored in a Google Drive folder. The Drive folder ID is supplied in the workflow's `instructions` (operator-written prompt). The agent reads the folder ID from its own prompt context and passes it to the tools. No schema changes, no embeddings, no ingestion pipeline.

## Context

Current agent tools live in `src/mailpilot/agent/tools.py` and are wired via `AgentDeps` in `src/mailpilot/agent/invoke.py:37-46`:

- `send_email`, `reply_email`, `create_task`, `cancel_task`, `record_enrollment_outcome`, `disable_contact`, `list_enrollments`, `search_emails`, `read_contact`, `read_company`, `read_email`, `noop`.

There is nothing for document retrieval. Gmail send/receive, Pub/Sub triggering, inbound routing, and inbound auto-reply workflows already exist (`docs/email-flow.md`, `docs/adr-03-workflow-model.md`), so the missing capability is purely Drive read access plus the two agent tools that surface it.

### New agent tools

Both follow the existing pattern in `src/mailpilot/agent/tools.py`: typed signature, dependency-injected `drive_client`, dict return shape, error dicts on failure.

```python
def list_drive_markdown(
    drive_client: DriveClient,
    folder_id: str,
) -> list[dict[str, str]]:
    """List Markdown files in a Drive folder.

    Returns: [{"file_id": ..., "name": ...}, ...]
    Errors:  {"error": "drive_unavailable" | "not_found" | ..., "message": ...}
    """


def read_drive_markdown(
    drive_client: DriveClient,
    file_id: str,
) -> dict[str, str]:
    """Read the content of a Markdown file from Drive.

    Returns: {"name": ..., "content": ..., "web_view_link": ...}
    Errors:  {"error": "drive_unavailable" | "not_found" | ..., "message": ...}
    """
```

`list_drive_markdown` calls `files.list` with:

```
q = "mimeType='text/markdown' and parents in '<folder_id>' and trashed = false"
fields = "files(id, name)"
```

`read_drive_markdown` calls `files.get(fileId, alt=media)` and returns the body verbatim.

### Decline behaviour

[...]

PR #109, opened 2026-04-30
docs(adr): ADR-09 Google Drive Markdown knowledge base

kborovik:
## Summary

- Adds `docs/adr-09-knowledge-base-rag.md` (Status: **Proposed**) documenting the minimal Google Drive Markdown lookup design that backs the KB-grounded auto-reply agent capability promised by the public demo.
- Two agent tools only: `search_drive_markdown(query)` and `read_drive_markdown(file_id)`. Reads `.md` files from a per-account Drive folder via the Drive API.
- No PostgreSQL changes, no embeddings, no pgvector, no chunking, no in-house ingestion pipeline. Operators handle PDF/Docs to Markdown conversion externally and drop the result into the folder.

## Design highlights

- New nullable column `kb_drive_folder_id` on `account` is the only schema change. Accounts without a folder configured have no KB and the search tool returns an empty list.
- Cross-account read protection: `read_drive_markdown` verifies the requested file is under `account.kb_drive_folder_id` before returning content -- mirrors the account-scoping check already used by `read_email`.
- Decline behaviour is prompt-driven, not enforced by a tool. The agent satisfies the existing "must call at least one tool per run" invariant by calling `search_drive_markdown` (legitimately empty) then `reply_email` with a polite decline.
- Drive client reuses the same service-account / domain-wide-delegation auth as Gmail. One scope addition: `https://www.googleapis.com/auth/drive.readonly`.

## Alternatives considered

- **pgvector + embeddings + ingestion pipeline.** Rejected for v1 -- it solves a problem (semantic search across long-form sources at scale) the demo and current product scope do not have. Migration path stays open: additive vector tables and a second search tool, with the Drive lookup tools left in place during transition.
- **In-house PDF / Docs to Markdown conversion.** Rejected as out of scope. Markdown-only is a deliberate constraint.
- **Local file system source.** Rejected -- Drive is already the operator collaboration surface and per-account impersonation gives natural multi-tenant isolation.

## Scope

This PR ships the design only. Tool implementation, `kb_drive_folder_id` column, `DriveClient`, and `AgentDeps` wiring follow on issue #108.

Refs #108

## Test plan

- [ ] Read the ADR end-to-end and confirm the design is what we want before merging.
- [ ] Confirm the "Proposed" status is appropriate (flip to "Accepted" on merge if so).

### DIFF
--- a/docs/adr-09-knowledge-base-rag.md
+++ b/docs/adr-09-knowledge-base-rag.md
@@ -10,114 +10,118 @@ The public demo at https://lab5.ca/demo/ requires the inbound auto-reply agent t
 
 Existing capabilities cover the surrounding loop end-to-end: Pub/Sub triggering, inbound sync, routing, agent invocation with Pydantic AI tools, and Gmail reply (`docs/adr-01-gmail-api-integration.md`, `docs/adr-03-workflow-model.md`, `docs/adr-04-email-routing.md`, `docs/email-flow.md`). The only missing capability is a way for the agent to look up authoritative content during a run.
 
-The default industry pattern is RAG: ingest source documents into a vector store, embed at query time, retrieve top-k chunks, ground the answer. That introduces a vector database (pgvector or external), an embedding provider, an ingestion pipeline, chunking strategy, and re-index logic. For the demo and the foreseeable product scope -- dozens of short product specs per account -- this is more machinery than the problem requires.
+The default industry pattern is RAG: ingest source documents into a vector store, embed at query time, retrieve top-k chunks, ground the answer. That introduces a vector database, an embedding provider, an ingestion pipeline, chunking strategy, and re-index logic. For the demo and the foreseeable product scope -- a handful of short product specs per workflow -- this is more machinery than the problem requires.
 
 ## Decision
 
-The knowledge base is a Google Drive folder containing Markdown files. Two agent tools read it directly via the Drive API. No PostgreSQL tables, no embeddings, no vector index, no chunking, no in-house ingestion.
+The knowledge base is a Google Drive folder containing Markdown files. The folder ID is supplied in the workflow's `instructions` (the operator-written prompt body). Two agent tools list and read the folder via the Drive API. No PostgreSQL tables, no embeddings, no vector index, no chunking, no in-house ingestion, no schema changes.
 
-### Source of truth
+### Source of truth and configuration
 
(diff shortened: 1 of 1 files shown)


kborovik:
A few design notes that didn't fit cleanly in the ADR but are worth recording for the implementation PR:

- **Snippet cost**: `search_drive_markdown` returning a snippet means one extra `files.export` per hit. For a typical search returning 1-3 results this is fine; if it ever becomes a hot path, drop the snippet field and let the agent call `read_drive_markdown` on whichever hit looks most promising by name alone.
- **Drive `fullText contains` quirks**: the operator works on word stems for Western languages but does not do synonyms. The system prompt should explicitly tell the agent to retry with paraphrases if the first query returns empty -- otherwise the decline path will fire too eagerly.
- **Empty-folder vs unreachable-Drive**: both currently surface to the agent as "no hits". Worth distinguishing at implementation time -- a Drive 5xx should return an error dict (`{"error": "drive_unavailable", ...}`) so the agent replies with "I can't access our docs right now" rather than the canonical "we don't cover that" decline.
- **Folder ID storage**: putting `kb_drive_folder_id` on `account` keeps the change to one column and matches how Drive folders are operationally bound to accounts. If we ever want multiple KBs per account (e.g. product docs + sales playbook), this becomes a `kb_source` table -- straightforward additive migration.
- **Future semantic-search migration**: the Drive lookup tools and a future `search_indexed_drive_documents` tool can coexist. The system prompt would prefer the indexed tool and fall back to keyword Drive search, giving a graceful upgrade with no breakage.

kborovik:
Revised the ADR (commit dfd3ff6) per discussion. Key changes:

- **No `account` schema change.** Folder ID lives in the workflow's `instructions`, not on `account`. Zero migration.
- **Tools renamed and reshaped.** `search_drive_markdown` -> `list_drive_markdown(folder_id)`; `read_drive_markdown(file_id)` keeps its name but no longer does an implicit folder-membership check (it reads what the agent asks for).
- **Workflow-scoped KB falls out for free.** Different workflows on the same account can point at different folders, or share one, by what the operator writes into `instructions`.
- **Access control = Drive permissions.** The impersonated user must have view access to the folder; otherwise Drive returns 404 and the agent declines. Folder IDs in `instructions` are not secrets but also not access grants -- a reasonable boundary delegation.
- **Listing instead of `fullText contains`.** Folders are small enough that the LLM picks files by filename (and after reading, by content). Sidesteps Drive's keyword-search quirks.

Updated PR description and earlier insights comment are still mostly accurate; the only stale bit is the snippet discussion (no snippets in the new tool shape -- the agent reads the file directly).

kborovik:
Closing -- ADR is more weight than this small a change needs. Folding the design points directly into issue #108 instead.

PR #110, opened 2026-04-30
feat(agent): list_drive_markdown and read_drive_markdown tools for KB-grounded replies

### LINKED ISSUE #108: feat(agent): list_drive_markdown and read_drive_markdown tools for KB-grounded replies

kborovik:
## Summary

Adds `list_drive_markdown` / `read_drive_markdown` agent tools to let inbound auto-reply workflows ground answers in Markdown documents stored in a Google Drive folder, per issue #108.

The PR also completes SPEC tasks T11–T20: CLI JSON envelope standardization, `is_routed` on EmailSummary, operator_event routing to stderr, Pub/Sub notification span consolidation, explicit trigger arg on `invoke_workflow_agent`, single-invocation discipline for `enrollment run`, and the Scenario B rewrite for KB-grounded smoke tests.

## Drive KB tools

- `list_drive_markdown(folder_id)` -- enumerates `.md` files in a Drive folder.
- `read_drive_markdown(file_id)` -- returns `{name, content, web_view_link}`.

Folder ID lives in the workflow's `instructions` -- no schema change, no embeddings, no ingestion pipeline. The agent system prompt instructs list-then-read when a folder is referenced, and a polite decline when no file is relevant (decline still calls `list_drive_markdown` -> `reply_email`, so the "must call >=1 tool per run" invariant holds).

## Implementation notes

- `src/mailpilot/drive.py` -- new `DriveClient`, mirrors `GmailClient`. Single new scope `drive.readonly`. Per-account impersonation via `credentials.with_subject(account.email)`.
- `src/mailpilot/agent/tools.py` -- new tools; error dict on `HttpError` (`not_found` for 404, `drive_unavailable` otherwise).
- `src/mailpilot/agent/invoke.py` -- `AgentDeps` gains `drive_client`; tools registered; system prompt extended with KB grounding + decline behaviour; explicit `trigger` arg replaces implicit enum inference.
- `src/mailpilot/gmail.py` -- `resolve_credentials_path` promoted from private (now used by both `gmail.py` and `drive.py`).
- `src/mailpilot/cli.py` -- all `view`/`create`/`update` commands wrap response under `{"<singular>": {...}, "ok": true}` (SPEC §V13 envelope standardization, T20).
- `src/mailpilot/models.py` + `database.py` -- `is_routed` field added to `EmailSummary` and `list_emails` projection (T11).
- `src/mailpilot/operator_log.py` -- all `operator_event` output routed to stderr so stdout stays strict-JSON (T17).
- `src/mailpilot/pubsub.py` -- `pubsub.notification` span and `notification.received` log collapsed into single event (T13/T14).

Multi-tenant access control delegated to Drive's permission model: a workflow naming a folder the impersonated user cannot see receives 404 and falls into the decline path.

## Test plan

- [x] `tests/test_drive.py` -- DriveClient list/read, query shape, alt=media, utf-8 decode.
- [x] `tests/test_agent_tools.py` -- both new tools cover hit / no-hit / 404 / 5xx.
- [x] `tests/test_agent_invoke.py` -- patch sites updated so `DriveClient` construction is mocked alongside `GmailClient`.
- [x] `tests/test_cli.py` -- envelope shape verified for all updated commands.
- [x] `make check` clean (ruff + basedpyright strict + 682 pytest).
- [ ] Smoke-test Scenario B (live KB-grounded demo, requires operator-prepared Drive folder).

Resolves #108


kborovik:
Notes from the build:

- **Test fragility caught during T5.** Every `tests/test_agent_invoke*.py` site that previously patched only `mailpilot.agent.invoke.GmailClient` now also has to patch `DriveClient`, because `invoke_workflow_agent` constructs both. Locally the tests still passed without the new patch (the dev box has `google_application_credentials` configured), so this would have been a CI-only failure on a fresh checkout. Worth keeping in mind for the next dependency added to `AgentDeps`.
- **Decline path satisfies the tool-use invariant by accident.** The existing "must call >=1 tool per run" invariant in `agent/invoke.py` is preserved on the polite-decline path because the agent legitimately calls `list_drive_markdown` (legitimately surfacing no relevant file) before `reply_email`. If a future change ever lets the prompt say "decline without listing," the invariant will trip. Captured in `SPEC.md` §V5.
- **No retry decorator on `DriveClient`.** `GmailClient` wraps every call in `_retry_on_transient` with a logfire span; `DriveClient` does not. Per CLAUDE.md's YAGNI rule, KB reads are infrequent and a transient 5xx surfaces as `drive_unavailable` to the agent which can decline. If we start hitting Drive rate limits in practice, we can lift the gmail retry helper into a shared module.
- **`gmail.resolve_credentials_path` was promoted from private.** It is now used cross-module; the four `# pyright: ignore[reportPrivateUsage]` comments in `tests/test_gmail_utils.py` were removed in the same change. No behaviour change.
- **Markdown MIME filter is operator-visible.** `mimeType='text/markdown'` is what Drive assigns when you upload a `.md` file. If an operator drops a Google Doc or a `.txt` file into the KB folder, it will not be listed. Documented as the contract in `SPEC.md` §V3 and §C.

kborovik:
**Notable observations from the diff:**

- **T20 envelope standardization** is a breaking change for any parser that relied on inlined entity fields at the top level (e.g. `json["workflow_id"]` instead of `json["workflow"]["workflow_id"]`). The smoke-test SKILL.md now documents the correct extraction path via `§V13`, so future smoke-test sessions won't regress.

- **`operator_event` -> stderr** (T17) is a correctness fix, not just cleanup: single-shot commands like `enrollment run` pipe their stdout to parsers, and any stray line to stdout breaks JSON parsing. Routing to stderr ensures `journalctl` still captures the events while keeping stdout strict JSON.

- **`is_routed` on `EmailSummary`** (T11) exposes routing state in `email list` output -- this is the key field for diagnosing routing failures during smoke tests without needing a `view` round-trip per email.

- **`trigger` arg on `invoke_workflow_agent`** (T18) eliminates implicit enum inference from context. The previous approach was fragile -- a caller could inadvertently construct `AgentDeps` in a state that mapped to the wrong trigger. The explicit arg is a cleaner contract.

- **Drive access control via impersonation 404** is elegant: the system trusts Drive ACLs rather than duplicating access checks. The decline path is still exercised by a tool call, so the "must call >=1 tool" invariant holds -- this was a non-obvious correctness constraint worth preserving.

PR #126, opened 2026-05-15
docs(designs): strip company and contact to bare minimum

kborovik:
Strip `company` + `contact` to identity-only rows (T47) and inline notes in agent read tools (T48).

## Summary

`company` and `contact` now carry only the minimum columns needed for cold-email campaigns; research/persona/firmographics live in the existing append-only `note` table. A shared `load_contact_view` / `load_company_view` helper inlines recent notes into both CLI (`contact view`, `company view`) and agent (`read_contact`, `read_company`) surfaces -- same shape for both consumers. Smoke test Scenario A updated with a nonce-based personalization gate to verify the agent actually reads both entities before composing.

## Changes

- `company`: drop 11 cols; survivors: id, name, domain, created_at, updated_at
- `contact`: drop 8 cols + status enum; add nullable `disabled_reason TEXT` (NULL = active)
- Partial index `idx_contact_active ON contact(id) WHERE disabled_reason IS NULL`
- `company create` / `contact create` gain `--note STR`; atomic insert in same txn (§V.47)
- `contact list` drops `--status`/`--domain`; gains `--include-disabled`
- `load_contact_view` / `load_company_view` in `database.py` -- shared CLI + agent renderer
- Notes inlined full-body, DESC order, capped at 10; totals reflect actual row count
- Contact view inlines parent company notes when `company_id IS NOT NULL`
- `read_contact` / `read_company` return §V.15 error dict on not-found (was: `None`)
- `_BASE` agent protocol gains §V.53 directive: treat notes as personalization context
- Smoke test A1a: seed hex-nonce notes; gate A3 requires both tokens in email body
- Outbound fixture: instructions require both reads + verbatim Reference codes
- Smoke-test report is chat-only (no auto-write to disk); add `.claude/.gitignore`

## Test plan

- [x] T47: schema strip -- unit tests in `test_database.py`, `test_cli.py`, `test_models.py`
- [x] T48: notes-inlined view -- db-helper coverage (DESC, totals, cap, orphan, company-notes branch); agent-tool found/inline/not-found; CLI symmetry via `load_*_view` patches
- [x] `_BASE` directive regression guard (test_agent_tools.py)
- [ ] Smoke test Scenario A with A1a nonce personalization gate (requires warm accounts)


kborovik:
**Notable patterns in this diff:**

- `load_contact_view` / `load_company_view` as a shared renderer (database.py) is a clean DRY win -- CLI output and agent tool returns are now byte-identical in shape, so any future format change only touches one place.
- Nullable `disabled_reason TEXT` replacing a `status` enum + `status_reason` pair simplifies every query that filters active contacts to `WHERE disabled_reason IS NULL` -- no enum cast, no two-column join condition.
- `_INLINE_NOTES_CAP = 10` is a module constant in `database.py`; if prompt-window pressure is observed in Logfire traces (input_tokens climbing), this is the first knob to turn.
- The hex-nonce approach in smoke test A1a is worth keeping as a pattern elsewhere: a cryptographically distinct, deterministic token makes LLM context use falsifiable rather than heuristic.

**Follow-up to consider:**

- A1a's personalization gate depends on `read_contact`/`read_company` inlining notes -- the prerequisite is now shipped (T48). First smoke run after merge will confirm the full chain.

PR #143, opened 2026-06-22
feat: mailpilot-campaign-test skill + lab5.ca outbound workflow

kborovik:
## What

Add `/mailpilot-campaign-test`, a skill that simulates an outbound cold-email campaign against real CRM contacts, plus the lab5.ca outbound workflow it exercises.

The skill reads a campaign Markdown file (subject plus body with `{first_name}`, `{company}`, `{title}` placeholders), personalizes it for up to nine real contacts, runs the §V.42 outbound body lint, live-sends each rendered copy from `outbound@lab5.ca`, and confirms delivery. The goal is to simulate the campaign as close to real data as possible without emailing anyone real.

## Accounts and aliases

- **Source:** `outbound@lab5.ca` sends every message.
- **Targets:** nine aliases `inbound1@lab5.ca` through `inbound9@lab5.ca`, which all deliver into the `inbound@lab5.ca` mailbox. Each selected contact maps to one alias by sequence.
- **Delivery mailbox:** `inbound@lab5.ca` receives every alias and is synced to confirm delivery.

Each real contact supplies the personalization data (name, title, company, even its real email), but the envelope recipient is rewritten to its alias. The real contact address is never emailed. The run is capped at nine messages, one per alias.

## Safety

- Real contacts are the personalization data source only. No real contact address is ever a recipient.
- Live Gmail traffic goes from `outbound@lab5.ca` to the alias mailbox, at most nine messages.
- The skill never starts `mailpilot run`, so no auto-reply fires even though `inbound@lab5.ca` carries the inbound demo workflow.

## Changes

- Rename the skill `campaign-test` to `mailpilot-campaign-test`.
- Retarget from a single sink (`hello@lab5.ca`) to the nine inbound aliases, capped at nine sends.
- `select_contacts` assigns each real contact its alias; `verify_delivery` syncs `inbound@lab5.ca`; the report adds an Alias column.
- Rewrite the example campaign to the lab5.ca outreach message.
- Add the outbound workflow `workflows/outbound-lab5-llm-lookup-work.toml` (name: "Optimize lookup-heavy work with LLM systems"), grounded in the lab5.ca root-page message, and bump the workflows submodule pointer.

## Design

- The §V.42 lint reuses `_check_spec_table` imported from `mailpilot.agent.tools`, identical to the send path.
- Deterministic Python scripts in `scripts/` shell out to the `mailpilot` CLI and emit compact JSON, matching `mailpilot-reply-test`.
- The workflow uses the `outbound-general` template (tools: `read_contact`, `read_company`, `send_email`, `record_enrollment_outcome`). Instructions forbid fabricating figures for the prospect and pin the body to plain Markdown with a pipe table for spec rows.

## Verification

- `ruff format`, `ruff check`, and `py_compile` pass on all scripts.
- Alias mapping verified: contacts 1-9 map to `inbound1-9@lab5.ca`.
- Full data path run end to end against the local DB (36 real contacts): preflight parses the campaign and confirms Google credentials; `select_contacts` picks 9 and assigns aliases; `personalize` renders real names and company domains with a clean §V.42 lint. The generated preview shows the real-contact-to-alias mapping.
- `basedpyright` scopes `src/` and `tests/` only, so the skill scripts add no type-check burden.
- The full live send was not exercised because the `lab5.ca` test accounts and aliases are not in the local DB at build time; the send path reuses the exact `email send` invocation proven by `mailpilot-reply-test`.

## Note

The workflow definition was pushed to `kborovik/workflows` main per your instruction. Its `CLAUDE.md` treats landing there as a production promotion, so test the message with `/mailpilot-campaign-test` before importing the workflow against live contacts.


kborovik:
Added an Opus critique sub-agent to the skill (commit b77adb8).

**New phase 6 -- Critique (Opus sub-agent).** After delivery, one sub-agent reads each sent email with its contact and company context and scores it against a cold-email rubric: relevance and personalization, subject, value proposition, credibility, call to action, concision, tone, and spam risk. It returns per-email strengths, weaknesses, and the single highest-impact fix.

- `scripts/critique_prep.py` bundles each sent email with the recipient's contact fields and company profile (one `mailpilot company view` per domain, cached) into `critique_input.json`, so the sub-agent reads one file rather than running many queries.
- `references/marketing-rubric.md` holds the marketing standards (the grounding rule, eight scored dimensions, scoring scale, and required per-email output).
- `generate_report.py` folds in a per-email critique score column and a critique section. The critique is advisory: it never changes the PASS or FAIL verdict.
- `allowed-tools` gains `Agent`; the sub-agent does its own reading and returns only a short summary, so email bodies never enter the orchestrator window.

Verified: ruff, py_compile, `critique_prep` against the local DB (handles null company gracefully), profile extraction unit-tested with a mocked CLI response, and the report renders the critique column plus section.

PR #149, opened 2026-06-24
fix(campaign-test): guard accounts with positional account view

### LINKED ISSUE #147: campaign-test step 0b guard uses a nonexistent flag, account view --email
## What happened

The `/mailpilot-campaign-test` skill guards account creation in step 0b with `mailpilot account view --email <addr>`. The `account view` command has no `--email` option. It takes a positional `ACCOUNT_REF`. The guard command exits 2 every time, so the `||` always runs `account create`.

## Command

```
uv run mailpilot account view --email outbound@lab5.ca
```

## Error output

```
Usage: mailpilot account view [OPTIONS] ACCOUNT_REF
Try 'mailpilot account view --help' for help.

Error: No such option: --email Did you mean --help?
```

Exit code is 2.

## Expected

`account view` returns exit 0 for an account that exists, so the `||` skips `account create`. The correct form is positional:

```
uv run mailpilot account view outbound@lab5.ca   # exit 0 when the account exists
```

## Effect (corrected)

The guard never short-circuits. `account create` runs on every invocation. It stays harmless because `account create` is idempotent. `create_account` uses `INSERT ... ON CONFLICT (email) DO NOTHING` and returns `None` on a duplicate (§V.16). No `UniqueViolation` is raised. The CLI then emits a structured `duplicate_key` envelope on stderr and exits 1.

Correction to the original report: each run records two `duplicate_key` error envelopes on stderr, not two `account.create` exception spans. The `account.create` span closes clean. `cli_mutation` catches the controlled `output_error` `SystemExit`, holds it, and re-raises it after the span closes, so the span carries no escaping exception (§V.54, §B.107). `logfire.exception` never fires for this path.

The real cost is harmless stderr noise per run, not a Logfire error.

## Fix (not applied)

The positional form is correct: replace `account view --email <addr>` with `account view <addr>` in step 0b of `.claude/skills/mailpilot-campaign-test/SKILL.md`. We chose not to apply it. The CLI handles the duplicate gracefully, so the guard noise is harmless.

Found during campaign-test run 6542df4e.

kborovik:
## What

Step 0b of the mailpilot-campaign-test skill guarded account creation with `mailpilot account view --email <addr>`. The `account view` command has no `--email` option. It takes a positional `ACCOUNT_REF`.

The bad flag made the guard exit 2 on every run, so the `||` fallback always ran `account create`. Each run recorded two `duplicate_key` errors and two `account.create` exception spans in Logfire. The guard would not protect a future non-idempotent create path.

## Change

Both `account view` guards now use the positional form. `account view <addr>` exits 0 for an existing account, so the `||` skips `account create`. The `account create --email ... --display-name ...` half of each line is unchanged; `--email` is correct for `account create`.

## Verification

- `account view --help` lists only `--help`; the positional `ACCOUNT_REF` is confirmed.
- `account view outbound@lab5.ca` exits 0 for an existing account; `account view --email outbound@lab5.ca` exits 2. Confirmed by hand.
- Repo-wide search (hidden directories included) finds the pattern only at the two fixed lines.

Closes #147

### DIFF
--- a/.claude/skills/mailpilot-campaign-test/SKILL.md
+++ b/.claude/skills/mailpilot-campaign-test/SKILL.md
@@ -152,8 +152,8 @@ Preflight (step 1) fails when `outbound@lab5.ca` or `inbound@lab5.ca` is absent,
 which is the state right after `make clean` wipes them. Create each account only
 if it is missing:
 ```bash
-uv run mailpilot account view --email outbound@lab5.ca >/dev/null 2>&1 || uv run mailpilot account create --email outbound@lab5.ca --display-name "MailPilot Outbound"
-uv run mailpilot account view --email inbound@lab5.ca  >/dev/null 2>&1 || uv run mailpilot account create --email inbound@lab5.ca  --display-name "MailPilot Inbound"
+uv run mailpilot account view outbound@lab5.ca >/dev/null 2>&1 || uv run mailpilot account create --email outbound@lab5.ca --display-name "MailPilot Outbound"
+uv run mailpilot account view inbound@lab5.ca  >/dev/null 2>&1 || uv run mailpilot account create --email inbound@lab5.ca  --display-name "MailPilot Inbound"
 ```
 The `account view` guard makes this idempotent: `account create` errors with
 `duplicate_key` on an existing email, so the create runs only when `view` reports


kborovik:
Closing unmerged. The SKILL.md guard bug is harmless: the CLI handles the duplicate `account create` gracefully (clean span, structured `duplicate_key` envelope, §V.54 / §B.107). No code change needed. See #147 for the corrected analysis.

